PR #46, opened 2014-12-29
Proper conversion to/from JS when issuing queries to Cassandra

d00rman:
Until now, values were only sporadically converted between rb-cassandra and Cassandra types. This PR resolves this, and systematically converts values from JS to Cassandra when issuing queries, and back to JS after fetching the result rows. It also introduces conversion schemes for set values whose underlying types need conversion (such as set<varint>, set<josn>, etc.).

This PR aims at resolving [T84900](https://phabricator.wikimedia.org/T84900)

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -434,9 +434,6 @@ dbu.buildPutQuery = function(req, keyspace, table, schema) {
     for (var key in req.attributes) {
         var val = req.attributes[key];
         if (val !== undefined && schema.attributes[key]) {
-            if (val && val.constructor === Object) {
-                val = JSON.stringify(val);
-            }
             if (!schema.iKeyMap[key]) {
                 keys.push(key);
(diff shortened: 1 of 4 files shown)


d00rman:
Hey guys, can we merge this?

lib/dbutils.js
```diff
@@ -270,6 +280,40 @@ function decodeTimestamp (ts) {
...
+
+/**
+ * Generates read/write conversion functions for set-typed attributes
+ *
+ * @param {Object} convObj the conversion object to use for individual values (from dbu.conversions)
+ * @returns {Object} an object with 'read' and 'write' attributes
+ */
+function generateSetConvertor (convObj) {
+    if (!convObj) {
+        return { write: identityConversion, read: identityConversion };
```
gwicke: @earldouglas @d00rman I was joking about the identity package ;)

PR #47, opened 2014-12-29
first pass on schema evolution on table creation for table that already exists

hardikj:


lib/db.js
```diff
@@ -600,77 +601,100 @@ DB.prototype.createTable = function (reverseDomain, req) {
...
+        var schemaInfo = dbu.makeSchemaInfo(schema);
 
-    if (req.options) {
-        if (req.options.durability === 'low') {
-            replicationOptions = "{ 'class': 'NetworkTopologyStrategy', 'datacenter1': 1 }";
+        if (currentSchema) {
+            try {
+                assert.deepEqual(currentSchema, schemaInfo);
+            } catch(e) {
+                throw new Error('The existing schema does not match the new defination', currentSchema, schemaInfo);
```
gwicke: Typo. Also, I think we discussed a better wording before.

We should also make sure that we return a 4xx response, as the issue is really on the client side.

gwicke:
@hardikj, this needs a rebase.

lib/db.js
hardikj: >  We should also make sure that we return a 4xx response
I wondering where should I handle the 4xx

lib/db.js
```diff
@@ -610,77 +611,106 @@ DB.prototype.createTable = function (reverseDomain, req) {
         consistency = cass.types.consistencies[req.consistency];
     }
 
-    // Validate and normalize the schema
-    var schema = dbu.validateAndNormalizeSchema(req);
-
-    var schemaInfo = dbu.makeSchemaInfo(schema);
+    return this._getSchema(keyspace, consistency)
+    .then(function(gSchema) {
```
gwicke: Why is this `.then()` needed?

lib/db.js
```diff
@@ -610,77 +611,106 @@ DB.prototype.createTable = function (reverseDomain, req) {
...
-    //
-    // Always use NetworkTopologyStrategy with default 'datacenter1' for easy
-    // extension to cross-DC replication later.
-    var replicationOptions = "{ 'class': 'NetworkTopologyStrategy', 'datacenter1': 3 }";
+        var schemaInfo = dbu.makeSchemaInfo(schema);
 
-    if (req.options) {
-        if (req.options.durability === 'low') {
-            replicationOptions = "{ 'class': 'NetworkTopologyStrategy', 'datacenter1': 1 }";
+        if (currentSchema) {
```
gwicke: Maybe call this `currentSchemaInfo` for consistency?

lib/dbutils.js
```diff
@@ -20,6 +21,26 @@ var dbu = {};
  * # Section 1: low-level helpers
  */
 
+
+/*
+ * Error instance wrapping HTTP error responses
+ *
+ * Has the same properties as the original response.
+ */
+function HTTPError(response) {
```
gwicke: Maybe we should actually move this out to its own package (http-types or the like?), so that we can share it across preq, rbUtils & dbutils.

gwicke:
Looks good to me in general, a couple of very minor comments inline.

lib/dbutils.js
hardikj: I will handle this in another pull request

hardikj:
@gwicke updated this

index.js
```diff
@@ -56,6 +56,12 @@ RBCassandra.prototype.createTable = function (rb, req) {
         };
     })
     .catch(function(e) {
+        if (e.status >= 400) {
```
gwicke: Could you simply return e here?

lib/db.js
```diff
@@ -587,77 +588,102 @@ DB.prototype.createTable = function (reverseDomain, req) {
...
 
-    // Validate and normalize the schema
-    var schema = dbu.validateAndNormalizeSchema(req);
-
-    var schemaInfo = dbu.makeSchemaInfo(schema);
-
-    // console.log(JSON.stringify(internalSchema, null, 2));
+    return this._getSchema(keyspace, consistency)
+    .catch(function(err) {
+        // Donot throw error if table doesnot exist yet.
```
gwicke: Spelling: Don't

lib/db.js
```diff
@@ -587,77 +588,102 @@ DB.prototype.createTable = function (reverseDomain, req) {
...
+        // 8704 is a Error code for "Table does not exist"
+        if (err instanceof Object && err.code === 8704) {
```
gwicke: It would be better to catch this in _getSchema, and convert it into a regular 404 HTTPError.

gwicke:
@hardikj, could you have a look at the follow-up comments?

PR #51, opened 2015-01-03
Improve code coverage by updating tests

hardikj:


hardikj:
do we really need this peace of code? https://coveralls.io/files/410949538#L91

d00rman:
Great work for the GET/PUT on non-existent tables, @hardikj ! Speaking of that, it makes me think. Doing a get or a put on such tables shouldn't return a 500 status code. It's actually a client (4xx) 'problem'.
BTW, is there a reason for renaming simpleTable into simple-table?

hardikj:
> Doing a get or a put on such tables shouldn't return a 500 status code. It's actually a client (4xx) 'problem'.

Ah! very well spotted, but I think its restbase that convert those 500 to 4xx since the response first goes through restbase before going to client.

>  is there a reason for renaming simpleTable into simple-table
Yes, To cover this regex case here, https://coveralls.io/files/410978712#L84

d00rman:
> Ah! very well spotted, but I think its restbase that convert those 500 to 4xx since the response first goes through restbase before going to client.
Good point. Besides, restbase should be the only entity accessing the storage directly, so we may assume here that it is going to handle that case well. Well, let's sleep on it :)
> is there a reason for renaming simpleTable into simple-table
> Yes, To cover this regex case here, https://coveralls.io/files/410978712#L84
Ah, right, good find!

d00rman:
> do we really need this peace of code? https://coveralls.io/files/410949538#L91

Yes, we do. It allows us to get the schema info in the same format as the cached version.

hardikj:
>> do we really need this peace of code? https://coveralls.io/files/410949538#L91

> Yes, we do. It allows us to get the schema info in the same format as the cached version.

Reason, I asked that was that we are never reaching the then statement, I am not sure why?

gwicke:
@hardikj, great work!

PR #52, opened 2015-01-05
Add paging support (T85640)

hardikj:
This is still work in progress.
https://phabricator.wikimedia.org/T85640

Task list -

[ X ] convert  the byte array to token and append that to next url 
[ X ] think on what happens if malicious users pass in random (or not so random) tokens.


earldouglas:
> earldouglas was assigned by gwicke just now

https://en.wikipedia.org/wiki/Hot_potato_%28game%29

hardikj:
seems like above test run failed due to coverall

PR #54, opened 2015-01-10
Use _del marker instead of straight deletion

hardikj:


hardikj:
@gwicke  what are your views on the comment https://github.com/wikimedia/restbase-cassandra/pull/54/files#diff-be228df81b99a384aa50ad50bde7fca0R578

do you want me  to just mark _del="time-uuid" for all indexes related to this table

d00rman:
@hardikj note that marking a row deleted is not as simple as it seems. First of all, you need to make sure this new behaviour does not conflict with the `dropTable` functionality, i.e. `dropTable` should not use `_delete` under the hood, since these are two distinct features. Next, do not forget that we encourage the usage of `UPDATE` on one and only one row in the table, meaning that attributes identifying the row must be present (unlike the current version where they are optional, implying a table drop). Finally, the tricky part are the index tables. You have to create a new record for the row being deleted, marking its `_del` column with the same timestamp.

Oh, yes, that also complicates retrieval (both of data and index keys). We have to be sure:
* not to return a row marked as deleted (AFAIK, this works properly already)
* pay attention to `_del` when retrieving info from index tables

Hope these notes help you :)

gwicke:
This looks good to me. Destructive deletion doesn't fit into our consistency / revisioning model. We filter on `_del` normally, although we should perform an audit (and make sure we have tests) to make sure that we do so everywhere.

d00rman:
> I am wondering though whether we should reuse `_put()` for this. We already enforce the presence of index entries there, and should make sure that index updates happen.

This is actually a great idea. Then, what needs to be done is:
* check the request parameters and compare them to the table's schema in order to ensure all of the index key fields have been provided (so as to be sure we are going to delete at most one row)
* forward the request to `_put()` setting `_del` appropriately

We still need to check the marker during retrieval, but since there is already a check for that in `_get()`, that might be it.

@hardikj feel like finishing this up? :)

gwicke:
The first bullet (enforcing index keys to be defined) is actually [already done](https://github.com/wikimedia/restbase-cassandra/blob/master/lib/dbutils.js#L502) when using `_put()`.

hardikj:
Yeah, seem like like just handling it to _put() would be enough!

gwicke:
LGTM. It looks like we need some tests around this.

PR #55, opened 2015-01-15
WIP: Convert restbase-cassandra into a new-style module

gwicke:
Remove the inline spec, and export operations only. This is intended to be
used with
https://github.com/gwicke/restbase/commit/fce27d997d4cfc46843b0b321dd2916bc8c0625b.

TODO: 
- Figure out a way to run the tests with spec & module
  - Distribute the spec & tests separately as `restbase-spec-sys-table`?
- Actually change the npm package name to `restbase-mod-table-cassandra`

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,45 +22,10 @@ function RBCassandra (options) {
     this.setup = this.setup.bind(this);
     this.store = null;
     this.handler = {
-        paths: {
-            // Table creation
-            '/v1/{domain}/{table}': {
-                // table creation
-                put: {
-                    request_handler: this.createTable.bind(this)
-                },
-                // delete
-                delete: {
-                    request_handler: this.dropTable.bind(this)
-                }
-            },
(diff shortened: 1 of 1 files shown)


d00rman:
> Figure out a way to run the tests with spec & module

Yep, this is something we need to figure out. With the separation of the interface from the module, the question is how to test these two, and can these be tested independently? The module ought to have its own tests, normally, but how to "test" the interface itself? Or, rather, to provide tests with the interface which should test the module, i.e. make sure it implements it.

> Actually change the npm package name to `restbase-mod-table-cassandra`

For now, I believe the wisest thing to do would be just to add a dependency to restbase called `restbase-mod-table-cassandra` which points to the topic branch on github. Then, once the transition is completed in restbase-cassandra, we can simply deprecate the "old" rb-cass package and activate the new one, since they will ultimately contain the same code.

PR #58, opened 2015-01-16
Export the module_info property

d00rman:
Patch the export object of restbase-mod-table-cassandra to export the module_info property as well (albeit not containing any useful info for now).

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ function RBCassandra (options) {
         module_info: {
             spec: null, // Re-export from spec module
             test: null, // Spec test function from spec module
-            dependencies: [],
+            dependencies: {},
             resources: [] // Dynamic resource dependencies, specific to implementation
         },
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -22,6 +22,12 @@ function RBCassandra (options) {
     this.setup = this.setup.bind(this);
     this.store = null;
     this.handler = {
+        module_info: {
+            spec: null, // Re-export from spec module
+            test: null, // Spec test function from spec module
+            dependencies: [],
```
gwicke: I think this should be an object if we plan to go with [the draft in key_value](https://github.com/wikimedia/restbase/blob/817cd5037d8d5b126deda3c1cdcdea0134c84d1f/mods/key_value.js#L371)

d00rman: Ah right, my bad. Will correct it now.

d00rman:
@gwicke fixed.

PR #66, opened 2015-02-11
Fix: Cover edge case of all provided attributes being null

d00rman:
When inserting into a table, we should use INSERT instead of an UPDATE
whenever all of the non-PK attributes are null, or are thought of as
being null by Cassandra, such as empty arrays and objects. This commit
fixes this issue and correctly determines which one to use.

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -29,6 +29,38 @@ dbu.arrayToSet = function arrayToSet(arr) {
     return o;
 };
 
+/**
+ * Checks if a value is null as interpreted by Cassandra.
+ * That includes null, undefined values as well as empty
+ * arrays, objects, strings and buffers.
+ *
+ * @param value the object/value to check
+ * @return boolean whether Cassandra treats the value as null
+ */
+dbu.isNull = function isNull(value) {
+    // if the value is null or undefined, it is not considered a value
+    if (value === null || value === undefined) {
+        return true;
+    }
+    // if the value is an empty array, it is not considered a value
+    if (Array.isArray(value) && !value.length) {
+        return true;
+    }
+    // if the value is an empty object, it is not considered a value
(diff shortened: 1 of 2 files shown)


d00rman:
@gwicke fixed. Now it is checked for in the first (and only) arg traversal. Also added checks for empty strings and buffers, which Cassandra also does not distinguish from null.

gwicke:
I can't quite reproduce the issue in cqlsh: https://gist.github.com/gwicke/a60e0fcad0b91bc6cfe7

Is this an issue with the node driver only, or were you able to reproduce it in cqlsh as well?

lib/dbutils.js
```diff
@@ -29,6 +29,38 @@ dbu.arrayToSet = function arrayToSet(arr) {
     return o;
 };
 
+/**
+ * Checks if a value is null as interpreted by Cassandra.
+ * That includes null, undefined values as well as empty
+ * arrays, objects, strings and buffers.
```
gwicke: I can reproduce this for empty sets, but not for empty strings and buffers.

gwicke:
@earldouglas, this LGTM to me, but could use some reviewing from you.

lib/dbutils.js
```diff
@@ -376,9 +392,7 @@ dbu.makeSchemaInfo = function makeSchemaInfo(schema) {
             // this is a set-typed attribute
             type = set_type[1];
             // generate the convertors only if the underlying type has them defined
-            if (dbu.conversions[type]) {
```
earldouglas: Is it ok to drop this guard?

gwicke: Yes, as generateSetConvertor (damn that spelling..) now explicitly returns a default converter object.

earldouglas: Ah, gotcha.  :+1:

PR #76, opened 2015-02-27
Cassandra driver v2 integration

gwicke:
The new version of [the datastax cassandra driver](https://github.com/datastax/nodejs-driver) brings several improvements:

- Built-in encoding & decoding of non-native types (previously handled by our own cassandra-codec package). This is done without binary modules, which now makes the cassandra backend pure JS only.
- App-level hartbeats and better error handling. This should hopefully fix the occasional hangs we saw when nodes went down.

Since this is a major upgrade we should test it thoroughly before deploying it in production. With the combined test coverage of this module & restbase I am fairly confident though that it is doing the right thing™.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -22,6 +22,7 @@ function DB (client, options) {
 
     // cache keyspace -> schema
     this.schemaCache = {};
+    this.keyspaceSchemaCache = {};
     this.keyspaceNameCache = {};
 
     /* Process the array of storage groups declared in the config */
@@ -52,6 +53,10 @@ DB.prototype._makeInternalRequest = function (domain, table, query, consistency)
         columnfamily: 'data',
         schema: this.schemaCache[cacheKey]
     });
+    if (!req.schema) {
+        // Share the schema across domains that map to the same keyspace
+        req.schema = this.keyspaceSchemaCache[req.keyspace];
+    }
     if (req.schema) {
         return P.resolve(req);
     } else {
@@ -74,6 +79,7 @@ DB.prototype._makeInternalRequest = function (domain, table, query, consistency)
                 var schema = JSON.parse(res.items[0].value);
                 self.keyspaceNameCache[cacheKey] = req.keyspace;
                 self.schemaCache[cacheKey] = req.schema = dbu.makeSchemaInfo(schema);
+                self.keyspaceSchemaCache[req.keyspace] = req.schema;
             }
             return req;
         }, function(err) {
@@ -198,7 +204,7 @@ DB.prototype.infoSchema = dbu.validateAndNormalizeSchema({
     secondaryIndexes: {}
 });
 
-DB.prototype.infoSchemaInfo = dbu.makeSchemaInfo(DB.prototype.infoSchema);
+DB.prototype.infoSchemaInfo = dbu.makeSchemaInfo(DB.prototype.infoSchema, true);
 
 DB.prototype.get = function (domain, query) {
     var self = this;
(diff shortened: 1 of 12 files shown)


d00rman:
Rebased on master

d00rman:
I have modified the tests a bit so that those that are correct, are reported as such. We still have two issues to tackle:

1. it seems secondary indices still need some work, since there is one [test not passing](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/cassandra_driver_v2/test/index.js#L689)
2. type conversion: now that the driver uses its own types for *varint*, *decimal* and *(time)uuid*, we would need to convert them rather carefully to some JS-native types, since exposing the driver types is not an option.

gwicke:
@d00rman, re native types and conversion: We can apply the trivial conversions (toString/toNumber) in our conversion framework. Conversions on the way in are now transparently handled by the driver.

The downside I see for doing this for each get / put is that internal requests inside the backend (like secondary index updates) might involve several conversions that way. We could maybe fix this later by pushing conversions up into the api layer.

d00rman:
> We could maybe fix this later by pushing conversions up into the api layer.

You mean into restbase? I wouldn't be up for that. Alternatively, we could wrap the `get` operations and do the conversion there on the way out. That way we are sure only external reqs get the converted data, while internal shims keep the driver's types. Just have to see if on the way out we can get our hands on the schema. Alternatively, we have to check if putting that in `db.get()` suffices. If I'm not mistaken, internal requests use `db._get()` only.

gwicke:
> You mean into restbase? 

No, into index.js or maybe the get/put wrappers, as you say.

lib/db.js
```diff
@@ -602,78 +603,116 @@ DB.prototype.createTable = function (reverseDomain, req) {
...
-    // TODO:2014-11-09:gwicke use info from system.{peers,local} to
-    // automatically set up DC replication
-    //
-    // Always use NetworkTopologyStrategy with default 'datacenter1' for easy
-    // extension to cross-DC replication later.
-    var localDc = this.conf.localDc;
-    var replicationOptions = "{ 'class': 'NetworkTopologyStrategy', '" + localDc + "': 3 }";
+    return this._getSchema(keyspace, consistency)
+    .catch(function(err) {
+        // Donot throw error if table doesnot exist yet.
```
atdt: does not

d00rman:
This repo's tests are now passing, but 5 restbase tests are failing. Haven't looked into it deeply yet, but it's clear we'll need more validation before merging this PR.

gwicke:
@d00rman, forgot to open a PR for the restbase test fix. Now at https://github.com/wikimedia/restbase/pull/200. It's a small fix in the log handling to account for fixed cassandra log handling.

lib/dbutils.js
```diff
@@ -392,12 +381,31 @@ function generateSetConvertor (convObj) {
     return res;
 }
 
+// Conversion factories. We create a function for each type so that it can be
+// compiled monomorphically.
+function toString() {
```
d00rman: Why have generators here? It seems to me a plain function returning `val.toString()` suffices. (idem for `toNumber()` below.)

gwicke: That'd be polymorphic, so would probably perform worse, which matters as these transformations are applied to each bit of data of a given type. 

From my understanding of how v8 works, mildly polymorphic functions compile to several guards to select the actual monomorphic code to use. There is a limit to the number of guards as well, so it might even fall back to low-performance polymorphic code. 

There is however also the possibility that it inlines the code or specializes the guards per call site, which case this wouldn't make a difference. We'd have to benchmark to find out.

PR #77, opened 2015-02-27
Couple of minor fixes

d00rman:


d00rman:
Self-merging as these are simple fixes for #75

PR #78, opened 2015-03-03
Add ability to share column families between domains

gwicke:
This patch prefixes a private `_domain` key element to each hash key, which in
turn lets us share a single CF for several domains. With this patch, all but
two listing tests (for titles and revisions) in restbase are passing. This is
the expected consequence of our strategy of using compound primary keys. The
ability to list hash keys can be restored later using secondary indexes.

Also:

- Introduced an `InternalRequest` class and a `_makeInternalRequest` method to
  set up and systematically handle request-related state. This cuts down the
  number of positional parameters drastically and replaces common boilerplate
  in `get()`, `put()`.

- Removed the now-dead `getSchema` and `_getSchema` methods. This
  functionality is now provided by `_makeInternalRequest`.

- Prepared for a removal of the `table` property in queries. Conceptually, the
  logical table is the resource to execute a query on, so should be part of
  the URI rather than the request body.

- Adjusted some tests to reflect the current limitation of no listings on hash
  keys. We could lift this limitation later, as described above.

- Added systematic private attribute stripping (all attributes starting with
  on an underscore) on the way out. This removes the previously-leaking `_del`
  attribute from the output.

- Commented out paging support and the single corresponding test. This code is
  not actually secure (clients can pass in arbitrary paging state), and needs
  an overhaul before we can re-enable paging. See
  https://phabricator.wikimedia.org/T85640.

- Commented out _getSecondaryIndex. This is not currently used, and needs an
  overhaul / rethink.

Next steps:

- Map groups of domains to internal storage domains in `dbu.keyspaceName()`.
  This will actually enable sharing.

Bug: https://phabricator.wikimedia.org/T91188

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -84,7 +84,7 @@ DB.prototype._makeInternalRequest = function (domain, table, query, consistency)
                 // Need to parse the JSON manually here as we are using the
                 // internal _get(), which doesn't apply transforms.
                 var schema = JSON.parse(res.items[0].value);
-                req.schema = dbu.makeSchemaInfo(schema);
+                self.schemaCache[schemaCacheKey] = req.schema = dbu.makeSchemaInfo(schema);
             }
             return req;
         }, function(err) {
--- a/lib/db.js
+++ b/lib/db.js
@@ -8,6 +8,7 @@ var dbu = require('./dbutils');
 var cassID = dbu.cassID;
 var secIndexes = require('./secondaryIndexes');
 
+
 function DB (client, options) {
     this.conf = options.conf;
     this.log = options.log;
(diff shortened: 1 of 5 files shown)


lib/db.js
```diff
@@ -23,6 +24,88 @@ function DB (client, options) {
...
+        domain: domain,
+        table: table,
+        query: query,
+        consistency: consistency,
+        columnfamily: 'data'
+    });
+    var schemaCacheKey = JSON.stringify([req.keyspace, domain]);
+    req.schema = this.schemaCache[schemaCacheKey];
+    if (req.schema) {
+        return Promise.resolve(req);
```
d00rman: This should be `P.resolve(req)`

lib/db.js
```diff
@@ -23,6 +24,88 @@ function DB (client, options) {
...
+            if (res.items.length) {
+                // Need to parse the JSON manually here as we are using the
+                // internal _get(), which doesn't apply transforms.
+                var schema = JSON.parse(res.items[0].value);
+                req.schema = dbu.makeSchemaInfo(schema);
+            }
+            return req;
+        }, function(err) {
+            // Check if the keyspace & meta column family exists
+            return self.client.execute_p('SELECT columnfamily_name FROM '
```
d00rman: Should we wrap these `client.*_p` calls with `P.resolve()` to make sure there are bluebird promises?

lib/db.js
```diff
@@ -40,102 +123,41 @@ DB.prototype.infoSchema = dbu.validateAndNormalizeSchema({
...
 };
 
-DB.prototype._get = function (keyspace, req, consistency, table, schema) {
+DB.prototype._get = function (req) {
     var self = this;
 
-    if (!table) {
-        table = 'data';
+    if (!req.schema) {
+        throw new Error("restbase-cassandra: No schema for " + req.keyspace
```
d00rman: It's called restbase-mod-table-cassandra now :wink:

lib/db.js
```diff
@@ -474,8 +478,8 @@ DB.prototype._rebuildIndexes = function (keyspace, req, schema, limit, indexes)
         var tidKey = schema.tid;
 
         // Build a new request for the main data table
-        var dataReq = {
-            table: req.table,
+        var dataQuery = {
+            table: req.query.table,
```
d00rman: Can be just `query.table`

lib/db.js
```diff
@@ -484,8 +488,8 @@ DB.prototype._rebuildIndexes = function (keyspace, req, schema, limit, indexes)
         // that's empty, the entire index (within the numerical limits) will be updated.
         schema.iKeys.forEach(function(att) {
             if (att !== tidKey) {
-                dataReq.attributes[att] = req.attributes[att];
-                dataReq.proj.push(att);
+                dataQuery.attributes[att] = req.query.attributes[att];
```
d00rman: Idem, `= query.attributes[att]` is enough.

lib/db.js
```diff
@@ -503,27 +507,28 @@ DB.prototype._rebuildIndexes = function (keyspace, req, schema, limit, indexes)
...
         }
 
         // XXX: handle the case where reqTid is not defined!
-        var reqTid = req.attributes[schema.tid];
+        var reqTid = query.attributes[schema.tid];
         var reqTime = dbu.tidNanoTime(reqTid);
 
         // Clone the query, and create le & gt variants
-        var newerDataReq = extend(true, {}, dataReq);
+        var newerDataQuery = extend(true, {}, dataQuery);
```
d00rman: Should be able to use `dataQuery.extend()` here (for readability)

PR #80, opened 2015-03-03
Storage groups

d00rman:
Use storage groups to determine domain to keyspace mappings.

Bug: [T91188](https://phabricator.wikimedia.org/T91188)

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -110,6 +110,80 @@ dbu.makeValidKey = function makeValidKey (key, length) {
 };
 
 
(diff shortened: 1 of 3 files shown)


lib/dbutils.js
```diff
@@ -110,6 +110,80 @@ dbu.makeValidKey = function makeValidKey (key, length) {
 };
 
 
+/* The array of storage groups declared in the config */
+dbu.storageGroups = [];
```
gwicke: This is moving beyond stateless utilities. Should we move this to db.js?

PR #81, opened 2015-03-03
Some more storage group tweaks

gwicke:


gwicke:
Merged, so that we can test on the test cluster & base work on this.

PR #82, opened 2015-03-03
Don't multiplex meta tables across domains

gwicke:
Meta tables track the table schema and potentially other table metadata, which
can't differ between domains. Thus, don't multiplex on domain for meta tables.

This avoids per-domain table creation & schema update attempts.


gwicke:
@eevans, please have a look!

PR #83, opened 2015-03-04
Change the log level when stating a CF is to be created

d00rman:


gwicke:
@d00rman, I actually think it's really important to know when tables were created, as it is a) rare and b) might be caused by a bug or storage issue. That's why I deliberately made it a warning rather than 'info' or any of the other levels that wouldn't make it into logstash in production.

This log is only emitted on actual creation, not if the table already exists.

d00rman:
> This log is only emitted on actual creation, not if the table already exists.

To be precise, when a worker *tries* to create a table. Maybe it would be more useful to log this only when the creation actually succeeds, so that we have one log entry per CF, not 6*32 of them

gwicke:
>> This log is only emitted on actual creation, not if the table already exists.

> To be precise, when a worker *tries* to create a table. Maybe it would be more useful to log this only when the creation actually succeeds, so that we have one log entry per CF, not 6*32 of them

You are referring to the case of starting many restbase workers on an empty db? I think there are several remedies for that case that we discussed before. I'd be in favor of starting workers sequentially (or at least one first, followed by the remaining ones), so that only one worker ends up creating the tables. The other workers won't log anything in that case, as the schemas will be identical. There is also no logging on restart (unless there is a bug, or a new storage group was added that requires table creation).

gwicke:
@d00rman, does the above sound like a good option to you?

d00rman:
> does the above sound like a good option to you?

Yup. Closing this one.

PR #86, opened 2015-03-06
also index _domain attribute of new columnfamilies

eevans:
Create Cassandra-native secondary indexes on the `_domain` attribute.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -884,7 +884,7 @@ DB.prototype._createTable = function (req, schema, columnfamily) {
     // Create a Cassandra-native secondary index on the _domain attribute, for
     // all tables except meta.
     if (columnfamily !== 'meta') {
-	tableCreation.then(function() {
+	return tableCreation.then(function() {
 	    var createIdx = 'CREATE INDEX IF NOT EXISTS ON '
 		+ cassID(req.keyspace) + '.' + cassID(columnfamily) + ' ("_domain")';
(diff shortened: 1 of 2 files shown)


lib/db.js
```diff
@@ -877,9 +877,23 @@ DB.prototype._createTable = function (req, schema, columnfamily) {
...
     // Execute the table creation query
-    return tasks.then(function() {
+    var tableCreation = tasks.then(function() {
         return self.client.execute_p(cql, [], {consistency: req.consistency});
     });
+
+    // Create a Cassandra-native secondary index on the _domain attribute, for
+    // all tables except meta.
+    if (columnfamily !== 'meta') {
+	tableCreation.then(function() {
```
gwicke: `return tableCreation.then`..

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.5.0",
+  "version": "0.5.1",
```
gwicke: Sorry, I already used 0.5.1, so you'll have to use 0.5.2.

lib/db.js
d00rman: Also, please note that we use spaces, not tabs for indentation. The whole block here is tab-indented. Disclaimer, I'm a big fan of tabs, but ... :P

PR #87, opened 2015-03-27
Add maintenance script to thin out renders in a key_rev_value table

gwicke:
Bug: https://phabricator.wikimedia.org/T94196

### DIFF
--- a/maintenance/thin_out_key_rev_value_data.js
+++ b/maintenance/thin_out_key_rev_value_data.js
@@ -1,0 +1,63 @@
+"use strict";
+var P = require('bluebird');
+var cassandra = P.promisifyAll(require('cassandra-driver'));
+
+if (!process.argv[3]) {
+    console.error('Usage: node ' + process.argv[1] + ' <host> <keyspace>');
+    process.exit(1);
+}
+
+var client = new cassandra.Client({
+    contactPoints: [process.argv[2]],
+    keyspace: process.argv[3]
+});
+
+var delay = 50; // One deletion every 50ms
+
+var lastKey;
+var lastKeyCount = 0;
+function processRow (row) {
+    // We include the revision in the key, so that we keep one render per
+    // revision.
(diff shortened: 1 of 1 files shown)


maintenance/thin_out_key_rev_value_data.js
```diff
@@ -0,0 +1,63 @@
...
+
+var client = new cassandra.Client({
+    contactPoints: [process.argv[2]],
+    keyspace: process.argv[3]
+});
+
+var delay = 50; // One deletion every 50ms
+
+var lastKey;
+var lastKeyCount = 0;
```
d00rman: `lastKeyCount` is not used anywhere so can be removed

maintenance/thin_out_key_rev_value_data.js
```diff
@@ -0,0 +1,63 @@
...
+function processRow (row) {
+    // We include the revision in the key, so that we keep one render per
+    // revision.
+    var key = JSON.stringify([row._domain, row.key, row.rev]);
+    if (key !== lastKey) {
+        //if (lastKeyCount > 10) {
```
d00rman: remove

maintenance/thin_out_key_rev_value_data.js
```diff
@@ -0,0 +1,63 @@
...
+        //    console.log(lastKeyCount + ':' + lastKey);
+        //}
+        lastKey = key;
+        lastKeyCount = 1;
+        //console.log(row);
```
d00rman: idem, remove

maintenance/thin_out_key_rev_value_data.js
gwicke: See #88.

gwicke:
> I'd prefer an implementation that paged by token(), tracked its state, and was resumable, but if understand the limitations of this approach and are satisfied, then +1.

#88 has the ability to resume, but uses the slightly more general `pageState` instead of a plain token().

PR #88, opened 2015-03-30
Update with changes from production

gwicke:
- Don't use the native driver's stream / forEach capability. They lack
  back-pressure and have less than perfect error handling, even with the
  custom RetryPolicy added here. For example, when encountering tombstone
  overwhelm we need to slow down to paging one item at a time in order to have
  a chance of making any progress at all.
- Print the pageState on error & support resuming by passing it in
- Drop delays.

### DIFF
--- a/maintenance/thin_out_key_rev_value_data.js
+++ b/maintenance/thin_out_key_rev_value_data.js
@@ -1,18 +1,55 @@
 "use strict";
+
+/**
+ * Simple script to delete all but the newest entry per revision in a
+ * key_rev_value table.
+ */
+
 var P = require('bluebird');
 var cassandra = P.promisifyAll(require('cassandra-driver'));
+var util = require('util');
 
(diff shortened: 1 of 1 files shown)


maintenance/thin_out_key_rev_value_data.js
```diff
@@ -1,63 +1,115 @@
...
     console.error('Usage: node ' + process.argv[1] + ' <host> <keyspace>');
     process.exit(1);
 }
 
-var client = new cassandra.Client({
-    contactPoints: [process.argv[2]],
-    keyspace: process.argv[3]
-});
+// A custom retry policy
+function AlwaysRetry () {}
```
d00rman: This class seems to be unused?

gwicke: Yes, it is. I have kept it for now in case we want to use it later here or elsewhere.

PR #89, opened 2015-03-31
Stop creating native Cassandra indexes on _domain

gwicke:
A bug in Cassandra 2.1 causes inserts in tables with native secondary indexes
on _domain to fail if the size of any column value exceeds a fairly low
threshold (16k IIRC). This is a problem especially on tables backing content
tables like those behind key_rev_value storage.

This patch stops creating those secondary indexes by default. This will break
the hacky revision and title listings in RESTBase, which we should disable or
replace with API requests.

Next steps in the backend:

- Automatically drop existing secondary indexes on `_domain`. Naming scheme is
  `<table_name>_<column_name>_idx_<counter>`, with the main difficulty being
  the counter value.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -877,23 +877,9 @@ DB.prototype._createTable = function (req, schema, columnfamily) {
     });
 
     // Execute the table creation query
-    var tableCreation = tasks.then(function() {
+    return tasks.then(function() {
         return self.client.execute_p(cql, [], {consistency: req.consistency});
     });
-
-    // Create a Cassandra-native secondary index on the _domain attribute, for
-    // all tables except meta.
-    if (columnfamily !== 'meta') {
-        return tableCreation.then(function() {
-            var createIdx = 'CREATE INDEX IF NOT EXISTS ON '
-                + cassID(req.keyspace) + '.' + cassID(columnfamily) + ' ("_domain")';
-            return self.client.execute_p(createIdx, [], {consistency: req.consistency});
-        });
-    }
-    else {
-        return tableCreation;
-    }
-
 };
 
 DB.prototype._createKeyspace = function (req, options) {
--- a/lib/db.js
+++ b/lib/db.js
@@ -700,10 +700,19 @@ DB.prototype.createTable = function (domain, query) {
 
         var newSchemaInfo = dbu.makeSchemaInfo(newSchema);
 
+
         if (currentSchemaInfo) {
             // Table already exists
             // Use JSON.stringify to avoid object equality on functions
             if (currentSchemaInfo.hash === newSchemaInfo.hash) {
+                if (!currentSchemaInfo._domainIndexDropped) {
+                    // Hacky flag to avoid dropping index on each request.
+                    // TODO: Properly keep track of internal schema versions!
+                    currentSchemaInfo._domainIndexDropped= true;
+                    // Asynchronously drop native secondary index on _domain column
+                    self._dropDomainIndex(req);
+                }
+
                 // all good & nothing to do.
                 return {
                     status: 201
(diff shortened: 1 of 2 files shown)


gwicke:
See also: https://phabricator.wikimedia.org/T94148

d00rman:
Unfortunately, this won't work. We were creating 2i indices without names, meaning they cannot be removed by referencing them by name. I put some logging around the code selecting and removing them, here's what I got:

```
SELECTING INDICES
Will drop INDEX -> "local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB"."null"
```

Also, the code deletes only the first found match, however both before and after starting RB, `cqlsh` tells me the number of indices stays the same (and is quite high):

```
cqlsh> select index_name, keyspace_name from system.schema_columns where keyspace_name = 'local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB';

 index_name | keyspace_name
------------+--------------------------------------------------
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
       null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB

(15 rows)
```

So, two conclusions:
- we had a problem whereby we created way too many 2i indices, it seems (there are not 15 tables in that keyspace)
- the code from this PR does not remove them (not even one)

Since in production the indices do have names and their number is limited, IMO it'd be better to manually delete them (just as we added them manually) or create a separate script doing that.

Unfortunately, there doesn't seem to exist a way to delete an index without a name (or at least I can't find one). Maybe deleting them from the `system` keyspace would suffice?

Either way, I'd rather not add any code in the back-end, but simply remove the index-creation portion and if there is a way to delete unnamed indices, then provide a script to do that.

lib/db.js
```diff
@@ -877,23 +886,24 @@ DB.prototype._createTable = function (req, schema, columnfamily) {
...
+DB.prototype._dropDomainIndex = function(req) {
+    var self = this;
+    var cql = "select index_name from system.schema_columns where keyspace_name = ? "
+        + " and columnfamily_name = ? and column_name = '_domain';";
+    return self.client.execute_p(cql, [req.keyspace, req.columnfamily], {prepare: true})
+    .then(function(res) {
+        if (res.rows.length) {
+            // drop the index
+            return self.client.execute_p('drop index if exists ' + cassID(req.keyspace)
+                    + '.' + cassID(res.rows[0].index_name));
```
d00rman: You need to search for the first non-null name, consider this:

```
cqlsh> select index_name, keyspace_name from system.schema_columns where keyspace_name = 'local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB';

 index_name         | keyspace_name
--------------------+--------------------------------------------------
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
 data__domain_idx_8 | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB
               null | local_group_test_T_parsoid_dataW4ULtxs1oMqJeY1GB

(15 rows)
```

By putting some debugging code around your code, I could confirm it tries to remove the `null` index, instead of `data__domain_idx_8`.

gwicke: I don't need to go through all columns as I'm selecting `_domain` only, but you are right that I don't need to perform the `drop index if exists` query if the index name is null.

PR #90, opened 2015-04-01
Remove native-secondary-index creation

d00rman:
Due to a bug in Cassandra 2.1, we cannot effectively use native secondary indices on domains.

This PR reverts #86 and represents an alternative to #89 .


d00rman:
cc @eevans @gwicke

PR #92, opened 2015-04-15
implement revision retention policy

eevans:
This branch:

 * implements a test case for a policy of `latest`
 * adds support for CQL generation of `USING TTL` queries
 * adds an async-safe utility method for performing arbitrarily large Cassandra queries, and paging through the results
 * adds `RevisionPolicyManager` (currently supports policy types of `all`, and `latest`), and wires it into a refactored `DB#_rebuildIndexes` (renamed `DB#_backgroundUpdates`).

See: https://phabricator.wikimedia.org/T94524

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ coverage/
 .coveralls.yml
 node_modules
 npm-debug.log
+.tern-project
+.tern-port
+*~
--- a/lib/db.js
+++ b/lib/db.js
@@ -6,6 +6,7 @@ var TimeUuid = cass.types.TimeUuid;
 var extend = require('extend');
 var dbu = require('./dbutils');
 var cassID = dbu.cassID;
+var revPolicy = require('./revisionPolicy');
 var secIndexes = require('./secondaryIndexes');
 
 
@@ -525,7 +526,7 @@ DB.prototype._put = function(req) {
     .then(function(result) {
         // Kick off asynchronous local index rebuild
         if (schema.secondaryIndexes) {
-            self._rebuildIndexes(req, 3)
+            self._backgroundUpdates(req, 3)
             .catch(function(err) {
                 self.log('error/cassandra/rebuildIndexes', err);
             });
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -1,4 +1,7 @@
 coverage/
 .coveralls.yml
 node_modules
-npm-debug.log
\ No newline at end of file
+npm-debug.log
+.tern-project
+.tern-port
+*~
```
gwicke: What is .tern about?

lib/db.js
```diff
@@ -539,135 +540,114 @@ DB.prototype._put = function(req) {
...
 };
 
-
 /*
- * Index update algorithm
+ * Post-put background updates
  *
- * look at sibling revisions to update the index with values that no longer match
+ * look at sibling revisions to update the index with values that no longer
+ * match, and apply the schema-specified revision retention policy.
```
gwicke: This sounds like it's the same thing, when it is really two separate updates. Maybe separate the description of index updates from the retention policy?

lib/db.js
```diff
@@ -539,135 +540,114 @@ DB.prototype._put = function(req) {
...
     var self = this;
     var schema = req.schema;
     var query = req.query;
-    if (!indexes) {
-        indexes = Object.keys(schema.secondaryIndexes);
+    indexes = indexes || Object.keys(schema.secondaryIndexes);
+
+    // If there are no indexes, and the retention policy is 'all' (i.e.
+    // there are no revisions to cull), then there is no need to go further.
+    if (!(indexes.length || schema.revisionRetentionPolicy.type !== 'all')) {
```
gwicke: This might be cleaner as `if (!indexes.length && schema.revisionRetentionPolicy.type === 'all')`

lib/db.js
```diff
@@ -945,4 +926,28 @@ InternalRequest.prototype.extend = function(opts) {
...
+
+/**
+ * Invokes handleRow on each of the child handlers with the supplied row.
+ *
+ * @param  {object} row; a row object.
+ * @return a promise that resolves when the constituent promises do.
+ */
+UpdateHandler.prototype.handleRow = function(row) {
+    var r = [];
+    for (var i = 0; i < this.handlers.length; i++) {
```
gwicke: Simpler:
```javascript
return P.all(this.handlers.map(function(handler) { 
     return handler.handleRow(row); 
}));
```

lib/dbutils.js
```diff
@@ -109,11 +110,61 @@ dbu.makeValidKey = function makeValidKey (key, length) {
...
+ * async callback handlers.
+ *
+ * @param   {object} cassandra-driver Client instance
+ * @param   {string} CQL query string
+ * @param    {array} CQL query params
+ * @param   {object} options map
+ * @param {function} function to invoke for each row result
+ */
+dbu.eachRow = function eachRow(client, query, params, options, handler) {
+    function processPage(pageState, times) {
```
gwicke: Maybe 'retries' instead of 'times' would be clearer?

lib/dbutils.js
```diff
@@ -159,6 +210,50 @@ dbu.validateIndexSchema = function validateIndexSchema(schema, index) {
     return index;
 };
 
+dbu.validateAndNormalizeRevPolicy = function validateAndNormalizeRevPolicy(schema) {
+    // FIXME: define as constants somewhere apropos
+    var minGcGrace = 10;
+    var maxGcGrace = 31536000;
```
gwicke: Do we need to limit the max?

lib/dbutils.js
```diff
@@ -159,6 +210,50 @@ dbu.validateIndexSchema = function validateIndexSchema(schema, index) {
...
+    var minKeep = 1;
+    var maxKeep = 1000000000;
+
+    var policy;
+
+    if (!schema.revisionRetentionPolicy) {
+        policy = { type: 'all' };
+    } else {
+        policy = schema.revisionRetentionPolicy;
+        for (var key in policy) {
```
gwicke: Better: 
```javascript
Object.keys(policy).forEach(function(key) {
}
```
This performs slightly better, and does not leak the 'key' variable out of the block scope.

lib/dbutils.js
```diff
@@ -159,6 +210,50 @@ dbu.validateIndexSchema = function validateIndexSchema(schema, index) {
...
+                if (val !== 'all' && val !== 'latest') {
+                    throw new Error('Invalid revision retention policy type ' + val);
+                }
+                break;
+            case 'grace_ttl':
+                // valid grace_ttls are between minGcGrace and maxGcGrace seconds
+                if (!validNumber(val, minGcGrace, maxGcGrace)) {
+                    throw new Error('Invalid grace_ttl value: ' + val);
+                }
+                policy.grace_ttl = parseInt(val);
```
gwicke: Shouldn't we require a number?

lib/dbutils.js
```diff
@@ -159,6 +210,50 @@ dbu.validateIndexSchema = function validateIndexSchema(schema, index) {
...
+                break;
+            case 'count':
+                // Valid counts are between minKeep and maxKeep
+                if (!validNumber(val, minKeep, maxKeep)) {
+                    throw new Error('Invalid count value: ' + val);
+                }
+                policy.count = parseInt(val);
```
gwicke: Same here.

lib/dbutils.js
```diff
@@ -656,14 +753,24 @@ dbu.buildPutQuery = function(req) {
 
     var using = '';
     var usingParams = [];
-    var usingTypeHints = [];
-    var usingParamsKeys = [];
+    var usingTypeHints = [];     // XXX: ???
+    var usingParamsKeys = [];    // XXX: ???
```
gwicke: I think these are indeed dead code.

lib/dbutils.js
```diff
@@ -656,14 +753,24 @@ dbu.buildPutQuery = function(req) {
...
         using = ' USING TIMESTAMP ? ';
         usingParams.push(cass.types.Long.fromNumber(Math.round(query.timestamp * 1000)));
         usingParamsKeys.push(null);
     }
 
+    if (req.ttl) {
+        if (using.length) {
+            using += 'AND TTL ? ';
+        } else {
+            using += ' USING TTL ? ';
```
gwicke: Eventually it might be slightly cleaner to accumulate all USING branches in an array, and then do
```javascript
if (usingBits.length) {
   using = ' USING ' + usingBits.join(' AND ');
}
```

lib/dbutils.js
```diff
@@ -109,11 +110,61 @@ dbu.makeValidKey = function makeValidKey (key, length) {
...
+ * to provide back-pressure during paging; This function can safely execute
+ * async callback handlers.
+ *
+ * @param   {object} cassandra-driver Client instance
+ * @param   {string} CQL query string
+ * @param    {array} CQL query params
+ * @param   {object} options map
+ * @param {function} function to invoke for each row result
+ */
+dbu.eachRow = function eachRow(client, query, params, options, handler) {
```
gwicke: How about maintaining a counter in eachRow, and then pass that in as the second (optional) parameter into the handler?

That would avoid the need for clients to maintain counter state.

lib/revisionPolicy.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {object} db; instance of DB
+ * @param {object} request; request to use as baseline
+ * @param {object} schema; the table schema
+ */
+function RevisionPolicyManager(db, request, schema) {
+    this.db = db;
+    this.request = request;
+    this.schema = schema;
+    this.policy = schema.revisionRetentionPolicy;
+    this.noop = this.policy.type === 'all' ? true : false;
```
gwicke: Equivalent: `this.noop = this.policy.type === 'all';`

lib/revisionPolicy.js
```diff
@@ -0,0 +1,57 @@
...
+    if (self.count < self.policy.count) {
+        self.count++;
+        return P.resolve();
+    }
+
+    var request = self.request.extend({ ttl: self.policy.grace_ttl });
+    Object.keys(request.query.attributes).forEach(function(key) {
+        request.query.attributes[key] = row[key];
+    });
+    request.query.timestamp = null;
```
gwicke: You could override this directly in the extend.

lib/revisionPolicy.js
```diff
@@ -0,0 +1,57 @@
...
+
+    var query = dbu.buildPutQuery(request);
```
gwicke: This overwrites the index keys, but not necessarily all the columns. Will that still remove the entire row once the TTL elapses?

test/revision_policies.js
```diff
@@ -0,0 +1,130 @@
...
+var makeClient = require('../lib/index');
+var TimeUuid = cass.types.TimeUuid;
+var yaml = require('js-yaml');
+
+describe('MVCC revision policies', function() {
+    var db;
+    before(function() {
+        return makeClient({
+            log: function(level, info) {
+                if (!/^info|verbose|debug|trace|warn/.test(level)) {
```
gwicke: Might be simpler to say which level you want to log ;)

test/revision_policies.js
```diff
@@ -0,0 +1,130 @@
...
+        })
+        .then(function() {
+            return db.createTable("domains_test", {
+                table: 'revPolicyLatestTest',
+                options: { durability: 'low' },
+                attributes: {
+                    title: 'string',
+                    rev: 'int',
+                    tid: 'timeuuid',
+                    comment: 'string'
```
gwicke: Might be good to also add a column that is not in any index or projection.

gwicke:
LGTM overall. A couple of small comments inline.

.gitignore
eevans: See: http://ternjs.net

It integrates with your editor to add support for:

* Autocompletion
* Function argument hints
* Type querying
* Navigating to the definition of things
* Automatic refactoring

It's pretty cool.

lib/dbutils.js
eevans: I'm not sure if we strictly need a limit, and so I'm fine with removing this.  My thoughts though were that given the purpose, anything beyond a certain point (months? years? I dunno.) would almost certainly be an error.  For example 9007199254740991 seconds is more than 200 thousand years after the current interglacial period ends. :)

lib/revisionPolicy.js
eevans: Maybe I'm missing something, `InternalRequest#extend` only works on the attributes of the request, so I'd have to specifiy the entirety of `query`.

.gitignore
gwicke: Interesting. Hadn't heard of that yet.

lib/revisionPolicy.js
gwicke: Yeah, you are right.

lib/db.js
```diff
@@ -945,4 +922,26 @@ InternalRequest.prototype.extend = function(opts) {
...
+}
+
+/**
+ * Invokes handleRow on each of the child handlers with the supplied row.
+ *
+ * @param  {object} row; a row object.
+ * @return a promise that resolves when the constituent promises do.
+ */
+UpdateHandler.prototype.handleRow = function(row) {
+    return P.all(this.handlers.map(function(handler) {
```
gwicke: You can actually directly use `P.map()`.

gwicke:
@eevans: LGTM, happy to merge after Travis has done its part.

PR #95, opened 2015-04-23
revision retention policy schema migration

eevans:
This makes it possible to migrate from one revision retention policy to another (or from none / the default).  More generally, it provides the framework by which additional schema migration can be implemented as well.

Note: In the absence of an explicitly defined schema version, the new default becomes 0.  This means that existing RESTBase installations will need to have their schema versions explicity set to 1, (that will be handled in a separate pull request against wikimedia/restbase).

See: https://phabricator.wikimedia.org/T96612

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,7 @@ function RBCassandra (options) {
         operations: {
             createTable: this.createTable.bind(this),
             dropTable: this.dropTable.bind(this),
-            getTable: this.getTable.bind(this),
+            getTableSchema: this.getTableSchema.bind(this),
             get: this.get.bind(this),
             put: this.put.bind(this)
         }
@@ -147,9 +147,9 @@ RBCassandra.prototype.dropTable = function (rb, req) {
     });
 };
 
-RBCassandra.prototype.getTable = function (rb, req) {
+RBCassandra.prototype.getTableSchema = function (rb, req) {
     var domain = req.params.domain;
-    return this.store.getTable(domain, req.params.table)
+    return this.store.getTableSchema(domain, req.params.table)
     .then(function(res) {
         return {
             status: 200,
--- a/index.js
+++ b/index.js
@@ -161,8 +161,8 @@ RBCassandra.prototype.getTableSchema = function (rb, req) {
         return {
             status: 500,
             body: {
-                type: 'table_query_error',
-                title: 'Internal error in Cassandra table storage backend',
+                type: 'schema_query_error',
+                title: 'Internal error querying table schema in Cassandra storage backend',
                 stack: e.stack,
                 err: e,
                 req: req
--- a/index.js
+++ b/index.js
@@ -22,6 +22,7 @@ function RBCassandra (options) {
         operations: {
             createTable: this.createTable.bind(this),
             dropTable: this.dropTable.bind(this),
+            getTable: this.getTable.bind(this),
             get: this.get.bind(this),
             put: this.put.bind(this)
         }
@@ -137,6 +138,30 @@ RBCassandra.prototype.dropTable = function (rb, req) {
             status: 500,
             body: {
                 type: 'delete_error',
+                title: 'Internal error in Cassandra table storage backend',
+                stack: e.stack,
+                err: e,
+                req: req
+            }
+        };
+    });
+};
+
+RBCassandra.prototype.getTable = function (rb, req) {
+    var domain = req.params.domain;
+    return this.store.getTable(domain, req.params.table)
+    .then(function(res) {
+        return {
+            status: 200,
+            headers: { etag: res.tid.toString() },
+            body: res.schema
+        };
+    })
+    .catch(function(e) {
+        return {
+            status: 500,
+            body: {
+                type: 'table_query_error',
                 title: 'Internal error in Cassandra table storage backend',
                 stack: e.stack,
                 err: e,
--- a/lib/db.js
+++ b/lib/db.js
@@ -21,11 +21,7 @@ function DB (client, options) {
     // cassandra client
     this.client = client;
 
-
-    // cache keyspace -> schema
-    this.schemaCache = {};
-    this.keyspaceSchemaCache = {};
-    this.keyspaceNameCache = {};
+    this._initSchemaCache();
 
     /* Process the array of storage groups declared in the config */
     this.storageGroups = this._buildStorageGroups(options.conf.storage_groups);
@@ -33,6 +29,12 @@ function DB (client, options) {
     this.storageGroupsCache = {};
 }
 
+DB.prototype._initSchemaCache = function() {
+    // cache keyspace -> schema
+    this.schemaCache = {};
+    this.keyspaceSchemaCache = {};
+    this.keyspaceNameCache = {};
+};
 
 /**
  * Set up internal request-related information and wrap it into an
@@ -723,8 +725,8 @@ DB.prototype.createTable = function (domain, query) {
                     });
                     return self._put(putReq)
                     .then(function() {
-                        var cacheKey = JSON.stringify([domain,query.table]);
-                        self.schemaCache[cacheKey] = newSchemaInfo;
+                        // Force a cache update on subsequent requests
+                        self._initSchemaCache();
                         return { status: 201 };
                     });
                 })
@@ -925,7 +927,7 @@ DB.prototype.dropTable = function (domain, table) {
             {consistency: this.defaultConsistency});
 };
 
-DB.prototype.getTable = function(domain, table) {
+DB.prototype.getTableSchema = function(domain, table) {
     var cacheKey = JSON.stringify([domain,table]);
     var req = new InternalRequest({
         domain: domain,
--- a/lib/db.js
+++ b/lib/db.js
@@ -896,6 +896,33 @@ DB.prototype.dropTable = function (domain, table) {
             {consistency: this.defaultConsistency});
 };
 
+DB.prototype.getTable = function(domain, table) {
+    var cacheKey = JSON.stringify([domain,table]);
+    var req = new InternalRequest({
+        domain: domain,
+        table: table,
+        keyspace: this.keyspaceNameCache[cacheKey] || this._keyspaceName(domain, table),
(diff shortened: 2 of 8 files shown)


lib/dbutils.js
```diff
@@ -110,6 +110,72 @@ dbu.makeValidKey = function makeValidKey (key, length) {
...
+ * JSON.stringify creating identical output for two or more matching
+ * objects with matching assignment order.  The docs would suggest this is
+ * an unsafe assumption, but in practice it seems to be true.
+ *
+ * @param  {object} schema; a schema object
+ * @return {string} the schema serialized to string
+ */
+dbu.makeSchemaHash = function makeSchemaHash(schema) {
+    // Create a copy with no references to the original
+    function clone(obj) {
```
d00rman: `extend` does the same thing, and does so faster then JSON conversions. Any specific reasons for using JSON here?

lib/dbutils.js
```diff
@@ -110,6 +110,72 @@ dbu.makeValidKey = function makeValidKey (key, length) {
...
+            }
+        }
+        return 0;
+    }
+
+    // Creates an object with the attributes assigned in sorted order; The
+    // expectation is that two normalized objects that are equal can be
+    // serialized to identical strings using JSON.stringify.
+    function normalize(obj) {
+        if ((obj === null) || (obj === undefined)) {
```
d00rman: It seems this function assumes an object is passed in, but judging from the code Arrays or even other stuff may be slipped in unnoticed and turned into objects. It would probably be safer to say here:

```javascript
if (typeof obj !== 'object')
```

lib/schemaMigration.js
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * @return a promise that resolves when the migration tasks are complete
+ */
+SchemaMigrator.prototype.migrate = function() {
+    return P.each(this.migrators, function(migrator) {
+        return migrator.migrate();
+    });
+};
+
+module.exports = {
```
d00rman: Could be simply:

```javascript
module.exports = SchemaMigrator;
```

lib/schemaMigration.js
```diff
@@ -0,0 +1,170 @@
...
+};
+
+/**
+ * Table name handler
+ */
+function Table(db, current, proposed) {
+    Unsupported.call(this, 'table', current, proposed);
+}
+
+Table.prototype = Object.create(Unsupported.prototype);
```
d00rman: Can be also:

```javascript
util.inherits(Table, Unsupported);
```

d00rman:
Nice one @eevans !

index.js
```diff
@@ -146,6 +147,30 @@ RBCassandra.prototype.dropTable = function (rb, req) {
     });
 };
 
+RBCassandra.prototype.getTable = function (rb, req) {
```
gwicke: Maybe make this 'getSchema'?

lib/db.js
```diff
@@ -694,14 +694,43 @@ DB.prototype.createTable = function (domain, query) {
...
+                            attributes: {
+                                key: 'schema',
+                                value: newSchema
+                            }
+                        }
+                    });
+                    return self._put(putReq)
+                    .then(function() {
+                        var cacheKey = JSON.stringify([domain,query.table]);
+                        self.schemaCache[cacheKey] = newSchemaInfo;
```
gwicke: I'm a bit concerned that stuffing the cache like this can hide bugs from tests. We need to detect a schema change in any case (reload on error?), so we probably don't even need to clear the cache here.

d00rman:
Hm, so, how about this situation: we do some changes (i.e. bump the schema version) and deploy it, but find out something's wrong (maybe has nothing to do with the schema itself). So, what happens then is:

- schema version is X
- we deploy a new version of RB, which bumps that to X+1
- the change is stored into Cassandra
- but, wait, something's wrong, roll back!
- we now what do re-deploy the previous version, but since the version stored in C* is X+1, we will not be able to start RB, as that version holds the schema version X

So, bottom line, should we keep the //monotonically-increasing version// requirement? It seems to me that as far as the retention policy it concerned, it doesn't really matter. The same is true for attributes: assuming we support their addition and removal, then removing and adding it (and vice-versa) is an idempotent operation (I do concentrate here on the schema solely and omit the fact that if you remove and then add a column the corresponding data is missing). In either case, the back-end will have to figure out whether the user is adding or removing something, which has really no relation to the schema version.

lib/db.js
gwicke: We just discussed this, and decided that just clearing the local cache (all or the key only) should be fine, as schema changes need to be backwards-compatible anyway for rolling deploys to work.

index.js
```diff
@@ -146,6 +147,30 @@ RBCassandra.prototype.dropTable = function (rb, req) {
...
+            headers: { etag: res.tid.toString() },
+            body: res.schema
+        };
+    })
+    .catch(function(e) {
+        return {
+            status: 500,
+            body: {
+                type: 'table_query_error',
+                title: 'Internal error in Cassandra table storage backend',
```
gwicke: (Minor): Could maybe make this slightly more specific, by mentioning getTableSchema or the like.

lib/dbutils.js
```diff
@@ -110,6 +110,67 @@ dbu.makeValidKey = function makeValidKey (key, length) {
...
+    // Creates an object with the attributes assigned in sorted order; The
+    // expectation is that two normalized objects that are equal can be
+    // serialized to identical strings using JSON.stringify.
+    function normalize(obj) {
+        if (!(obj instanceof Object)) {
+            return obj;
+        }
+        var copy = {};
+        Object.keys(obj).sort().forEach(function(key) {
+            if (obj[key] instanceof Array) {
```
gwicke: This would ignore differences in the order of index definitions, which wouldn't be great.

lib/dbutils.js
```diff
@@ -110,6 +110,67 @@ dbu.makeValidKey = function makeValidKey (key, length) {
     }
 };
 
+/**
+ * Create a schema hash string.
```
gwicke: Do you think we could use https://www.npmjs.com/package/json-stable-stringify instead?

lib/schemaMigration.js
```diff
@@ -0,0 +1,170 @@
...
+
+var dbu = require('./dbutils');
+var P = require('bluebird');
+var util = require('util');
+
+var hash = dbu.makeSchemaHash;
+
+/**
+ * Base migration handler for unsupported schema
+ */
```
gwicke: This feels very much like Java ;)

test/schema_migrations.js
```diff
@@ -0,0 +1,114 @@
...
+
+var assert = require('assert');
+var fs = require('fs');
+var makeClient = require('../lib/index');
+var router = require('./test_router.js');
+var yaml = require('js-yaml');
+
+var hash = JSON.stringify;
+
+function clone(obj) {
```
gwicke: Alternative for shallow cloning: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

test/schema_migrations.js
```diff
@@ -0,0 +1,114 @@
...
+        return router.request({
+            uri: '/restbase.cassandra.test.local/sys/table/testTable0',
+            method: 'DELETE',
+            body: {}
+        });
+    });
+
+    it('migrates revision retention policies', function() {
+        var newSchema = clone(testTable0);
+        newSchema.version = 1;
```
gwicke: It would be nice to test without an explicit schema version first, as that's what we currently have in production.

lib/schemaMigration.js
d00rman: Totally! Exactly my thought when I was reading the code :)))

eevans: What about it feels like Java? The use of inheritance?

test/schema_migrations.js
eevans: Isn't that ECMA 6? 

This could be replaced with `extend` as suggested by @d00rman (and as I've already done elsewhere).

test/schema_migrations.js
eevans: Auh, right.  That's the way it was meant to be, I think I forgot to do the inital put of the base schema in `before()`.

eevans: Sorry, the first test *is* without an explicit version.  The table is first written with `testTable0` (with no version attribute set), and then again with `newSchema`, which *does* have a version assigned.

Either way, it will be more obvious with the first put in `before()`, so that's what I'll do.

lib/dbutils.js
eevans: Maybe, but https://github.com/substack/json-stable-stringify/issues/14 would need to be fixed first.

test/schema_migrations.js
gwicke: This refers to https://github.com/wikimedia/restbase-mod-table-cassandra/pull/95/files#diff-3b58377c7047ab11edbff7ec5d8f283aR266, where the default is now explicitly set to version 0.

lib/dbutils.js
```diff
@@ -252,9 +263,7 @@ dbu.validateAndNormalizeRevPolicy = function validateAndNormalizeRevPolicy(schem
 
 dbu.validateAndNormalizeSchema = function validateAndNormalizeSchema(schema) {
     if (!schema.version) {
-        schema.version = 1;
-    } else if (schema.version !== 1) {
-        throw new Error("Schema version 1 expected, got " + schema.version);
+        schema.version = 0;
```
gwicke: If we move the default backwards we'll have to make sure that we'll properly deal with existing stored schemas at version 1.

lib/schemaMigration.js
gwicke: Discussed this on IRC. Yeah, something along [the kingdom of the nouns](steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html). It's fine, was just my first gut reaction when I saw it.

eevans: I never seen that before; It's an amusing read. :)

PR #97, opened 2015-04-30
Fix and reimplement paging

hardikj:


lib/db.js
```diff
@@ -233,6 +234,60 @@ DB.prototype._get = function (req) {
...
+        var pageSize = req.query.pageSize;
+        var rows = [];
+        var options = {
+            consistency: req.consistency,
+            fetchSize: pageSize? pageSize : maxLimit,
+            prepare: true
+        };
+
+        if (req.query.next) {
+            options.pageState = new Buffer(jwt.verify(req.query.next, this.conf.salt_key), 'base64');
```
gwicke: Should we move the validation to RB?

lib/db.js
```diff
@@ -234,8 +234,61 @@ DB.prototype._get = function (req) {
...
+        delete req.query.limit;
+        buildResult = dbu.buildGetQuery(req);
+
+        if (req.query.next) {
+            options.pageState = new Buffer( req.query.next, 'base64');
+        }
+
+        return new P(function(resolve, reject) {
+            try {
+                self.client.eachRow(buildResult.cql, buildResult.params, options,
```
gwicke: `eachRow` doesn't have any back-pressure, and doesn't seem to be needed here anyway. Instead, we can just add support for passing in the pageState in the regular get handling path, similar to https://github.com/wikimedia/restbase-mod-table-cassandra/blob/ea853a27b3c4592c5be4ec2b8470e1902433b9a8/maintenance/thin_out_key_rev_value_data.js#L79-L83

lib/db.js
```diff
@@ -234,8 +234,61 @@ DB.prototype._get = function (req) {
...
+                    function(n, result){
+                        dbu.convertRows(result, req.schema);
```
gwicke: This is already handled in index.js, for all results.

lib/db.js
hardikj: Yup, that makes sense

lib/db.js
```diff
@@ -234,10 +234,36 @@ DB.prototype._get = function (req) {
...
-    return self.client.execute_p(buildResult.cql, buildResult.params,
-            {consistency: req.consistency, prepare: true})
+    // Index queries are currently handled in buildGetQuery. See
+    // https://phabricator.wikimedia.org/T78722 for secondary index TODOs.
+    //if (req.index) {
+    //    return this._getSecondaryIndex(keyspace, req, consistency, table, buildResult);
+    //}
+
+    // Paging request:
+    var buildResult;
```
eevans: A minor nit, but how about in-lining declaration of `buildResult` during assignment below?

lib/db.js
```diff
@@ -234,10 +234,36 @@ DB.prototype._get = function (req) {
...
+    var options = {consistency: req.consistency, prepare: true};
+
+    if (req.query.limit) {
+        var rows = [];
```
eevans: Why declare rows here?

lib/db.js
```diff
@@ -234,10 +234,36 @@ DB.prototype._get = function (req) {
...
+        options = {
```
eevans: It would be better to add the new attribute (`fetchSize`), rather than reassigning the entire object, but... (see next comment)

lib/db.js
```diff
@@ -234,10 +234,36 @@ DB.prototype._get = function (req) {
...
+            consistency: req.consistency,
+            fetchSize: req.query.limit,
+            prepare: true
+        };
+
+        // Delete limit from query 
+        // This is done to make buildGetQuery avoid appending of 'limit' in cql
+        delete req.query.limit;
```
eevans: `fetchSize` is used internally by the driver for paging through the requested results; why not use `limit` here?

lib/db.js
```diff
@@ -248,69 +274,17 @@ DB.prototype._get = function (req) {
                 length--;
             }
         }
+        if (result.meta && result.meta.pageState) {
+            var token = result.meta.pageState.toString('base64');
+            return {
+                items: rows,
+                next: token
+            };
+        }
```
eevans: This would be better as an if/else here, I think.

lib/db.js
gwicke: To me it seems to make more sense to generally translate limit to fetchSize, and remove the LIMIT handling in buildGetQuery. That also avoids the need to delete attributes here.

eevans:
LGTM :+1:

PR #98, opened 2015-05-01
schema attribute addition/removal

eevans:
Adds an implementation of schemaMigration.Attributes (which previously
inherited Unsupported).  Iteratatively adds and/or removes the delta
attributes of two schemas.

Bug: T97073

### DIFF
--- a/lib/schemaMigration.js
+++ b/lib/schemaMigration.js
@@ -1,10 +1,11 @@
 "use strict";
 
+require('core-js/shim');
+
 var dbu = require('./dbutils');
 var P = require('bluebird');
 var util = require('util');
 
-var hash = dbu.makeSchemaHash;
 
 /**
  * Base migration handler for unsupported schema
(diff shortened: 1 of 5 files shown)


lib/schemaMigration.js
```diff
@@ -1,10 +1,11 @@
 "use strict";
 
+require('core-js/shim');
```
gwicke: This is normally taken care of by the entry points, so that we don't have to re-do it in every library.

lib/schemaMigration.js
```diff
@@ -37,7 +38,7 @@ util.inherits(Table, Unsupported);
 /**
  * options object migration handler
  */
-function Options(db, current, proposed) {
+function Options(parent, current, proposed) {
```
gwicke: I wasn't quite sure what the 'parent' is in this case, and which interface it implements. If the intention is that migrators can be nested, then maybe call it 'parentMigrator'?

PR #100, opened 2015-05-06
back-end versioning

eevans:
Given the case where *both* a backend version migration, and a table
storage schema migration are needed, it stands to reason that the latter
might depend on the former.  Because of this, the two cases need to be
treated separately, and table schema migration must occur after any backend
migration has completed.

This changeset introduces a `_backend_version` attribute to the schema blob
but does not consider it when generating the hash.  DB#_createTable
first evaluates `_backend_version`, performs a back-end migration if
necessary, then compares hashes and performs a table schema migration if
necessary.  Finally, if any changes were made, the persisted schema blob
is updated.

Bug: T97845

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -9,7 +9,6 @@ var cassID = dbu.cassID;
 var revPolicy = require('./revisionPolicy');
 var SchemaMigrator = require('./schemaMigration');
 var secIndexes = require('./secondaryIndexes');
-var validator = require('restbase-mod-table-spec').validator;
 
 
 function DB (client, options) {
@@ -195,7 +194,7 @@ DB.prototype._resolveStorageGroup = function (domain) {
 
 
 // Info table schema
-DB.prototype.infoSchema = validator.validateAndNormalizeSchema({
+DB.prototype.infoSchema = dbu.validateAndNormalizeSchema({
     table: 'meta',
     attributes: {
         key: 'string',
@@ -667,6 +666,26 @@ DB.prototype._delete = function (req) {
     return this._put(req);
 };
 
+/**
+ * Evaluate, and if neccessary, perform a migration from one back-end version
+ * to another.
+ *
+ * @param  {object} req;  the current request object
+ * @param  {object} from; schema info object representing current state
+ * @param  {object} to;   schema info object representing proposed state
+ * @return {boolean} a promise that resolves to true if a back-end migration
+ *         occurred.
+ */
+DB.prototype._migrateBackend = function(req, from, to) {
+    // Perform a backend migration, as-needed.
+    switch(from._backend_version) {
+    case 0:
+        return this._dropDomainIndex(req);
+    default:
+        return P.resolve();
+    }
+};
+
 DB.prototype.createTable = function (domain, query) {
     var self = this;
     if (!query.table) {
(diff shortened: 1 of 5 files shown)


lib/db.js
```diff
@@ -678,63 +696,74 @@ DB.prototype.createTable = function (domain, query) {
...
-                        status: 400,
-                        body: {
-                            type: 'bad_request',
-                            title: 'The table already exists, and its schema cannot be upgraded to the requested schema ('+error+').',
-                            keyspace: req.keyspace,
-                            schema: newSchema
-                        }
-                    });
+            // Carry out any back-end migration (if needed).
+            return self._migrateBackend(req, currentSchemaInfo, newSchemaInfo)
```
gwicke: The hash check is normally the fast and common path on startup, as it is executed over a dozen times per worker per startup. I think it's important to still check that before performing any migrations.

lib/db.js
```diff
@@ -678,63 +696,74 @@ DB.prototype.createTable = function (domain, query) {
...
-                    var putReq = req.extend({
-                        columnfamily: 'meta',
-                        schema: self.infoSchemaInfo,
-                        query: {
-                            attributes: {
-                                key: 'schema',
-                                value: newSchema
+                             
+                // Use JSON.stringify to avoid object equality on functions
+                if (currentSchemaInfo.hash === newSchemaInfo.hash) {
```
gwicke: This should probably be checked before the backend is migrated.

lib/dbutils.js
```diff
@@ -265,6 +274,9 @@ dbu.validateAndNormalizeSchema = function validateAndNormalizeSchema(schema) {
     if (!schema.version) {
         schema.version = 1;
     }
+    if (!schema._backend_version) {
```
gwicke: Is the intention to upgrade from both 0 and `undefined` to 1 here, or do you only want to upgrade from `undefined`?

lib/db.js
eevans: Here, the hash pertains to everything *except* the backend (the backend version is excluded from the hash), so the backend is first checked independently (in what is now called `_maybeMigrateBackend`), and a backend migration is performed (if needed), whether a normal migration is required or not.  After a successful test of, (and possible migration of) the backend, the table schema is tested, and a migration of it is performed (again, if needed).

TL;DR I think the logic is correct, even if it doesn't read so clearly.  Suggestions on how to address that are welcome, of course.

gwicke: The point is partly about performance. The hash comparison is quick, which matters as it is performed hundreds of times per worker in the production setup, with each type of table tested a dozen times or so.

Without benchmarking I can't say for sure whether performing the backend migration every time is adding too much overhead. It does not seem to be necessary to perform it on each request though, if we can avoid doing so by including the backend version in the hash.

lib/db.js
```diff
@@ -646,37 +666,53 @@ DB.prototype.createTable = function (domain, query) {
...
-                if (!currentSchemaInfo._domainIndexDropped) {
-                    // Hacky flag to avoid dropping index on each request.
-                    // TODO: Properly keep track of internal schema versions!
-                    currentSchemaInfo._domainIndexDropped= true;
-                    // Asynchronously drop native secondary index on _domain column
-                    self._dropDomainIndex(req);
+            var migrationPromise = P.resolve();
+            var emptyPromise = migrationPromise;
+            // First carry out any back-end migration (if needed).
+            if (currentSchemaInfo._backend_version !== newSchemaInfo._backend_version) {
```
gwicke: I'm wondering why you didn't include the backend version
in the hash
15:26:56 +gwicke you could then move the backend
migration inside the hash conditional
15:28:37 +gwicke it sounds like all it would take to do
that would be to remove https://github.com/eevans/restbase-mod-table-cassandra/blob/backend-vers/lib/dbutils.js#L125-L126

gwicke:
@eevans, overall this LGTM, only nit is the comment (see above).

gwicke:
@eevans @d00rman: Lets perhaps wait once more until after the next deploy to avoid pushing out too many things at once, and then schedule just this change?

d00rman:
I agree, let's avoid shooting ourselves in the foot.

gwicke:
Thi has bit-rotted since (tests are failing).

gwicke:
@d00rman @Pchelolo: Could you review this?

lib/dbutils.js
```diff
@@ -297,6 +306,9 @@ dbu.validateAndNormalizeSchema = function validateAndNormalizeSchema(schema) {
...
+        schema._backend_version = dbu.CURRENT_BACKEND_VERSION;
```
Pchelolo: @eevans I think this all requires a rebase, because `validateAndNormalizeSchema` have been moved to the spec module, so this definitely wouldn't merge

Pchelolo:
LGTM

PR #104, opened 2015-05-07
Be robust against schema update races

gwicke:
Don't fail if columns to add already exist, or columns to remove don't exist
any more. This make sure that racing workers eventually make progress with
schema updates.


eevans:
:+1:

PR #107, opened 2015-05-13
Increase the max schema agreement wait time

gwicke:
Task: https://phabricator.wikimedia.org/T98934


eevans:
FWIW, I doubt this is going to help, (it probably won't hurt either).

PR #109, opened 2015-06-22
move schema validation tests to dbutils

eevans:
While being related to revision retention policy, the affected tests
test validation of the schema, and are thus better treated as unit
tests in `test/dbutils.js`.

This is also necessary to break the tests dependency on dbutils,
(relevant to T103406).

Bug: T103406

### DIFF
--- a/test/dbutils.js
+++ b/test/dbutils.js
@@ -67,3 +67,44 @@ describe('DB utilities', function() {
             dbu.makeSchemaHash(testTable0b));
     });
 });
+
+describe('Schema validation', function() {
+    it('rejects invalid revision retention policy schema', function() {
+        var policies = [
+            {
+                type: 'bogus'    // Invalid
+            },
+            {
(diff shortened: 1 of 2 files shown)


test/dbutils.js
```diff
@@ -66,4 +66,43 @@ describe('DB utilities', function() {
             dbu.makeSchemaHash(testTable0a),
             dbu.makeSchemaHash(testTable0b));
     });
+
+    it('validation rejects invalid revision retention policy schema', function() {
```
d00rman: since both tests are validation tests, you could wrap these two inside another `describe()` section (nested in the //DB utilities// one)

d00rman:
Minor nit, LGTM otherwise

gwicke:
LGTM overall as well.

test/dbutils.js
eevans: Yeah, good call.

PR #110, opened 2015-06-23
cleanup and refactor tests

hardikj:
https://phabricator.wikimedia.org/T99696


gwicke:
Uh oh. This looks like it might conflict with @eevan's work on moving tests to a shared module, for use by all table storage backends.

/cc @d00rman

hardikj:
Oh, thats ok, I can close it if @eevans have already started with the implementation.
Else, we can just move this code to the new module and start building up on it?

eevans:
> Else, we can just move this code to the new module and start building up on it?

I think it's OK to move forward; The tests will need to be refactored a little post-move, and I haven't start doing that yet, so let just proceed as-is.

gwicke:
@eevans, could you review this PR?

eevans:
> @eevans, could you review this PR?

Yeah, I'm working on it (sorry, should have made that clear).

test/mocha.opts
```diff
@@ -0,0 +1 @@
+--recursive
```
eevans: With `--recursive` enabled, I get failures because `test/dump` is loaded to search for tests, and the module `libxmljs` is missing.  Could you move `test/dump` to `scripts/`, or `maintenance/`?

/cc @gwicke @d00rman in case they want to bikeshed the exact destination :)

test/index.js
```diff
@@ -1,1271 +1,8 @@
...
-                    uri: "/restbase.cassandra.test.local/sys/table/unversionedSecondaryIndexTable",
-                    method: "delete",
-                    body: {}
-                });
-            });
-        });
-    });
-});
+// Run jshint as part of normal testing
+require('mocha-jshint')();
```
eevans: This is OK I think, but at least remove `utils/jshint.spec.js`, these two files are now redundant.

test/functional/simeple.js
```diff
@@ -0,0 +1,380 @@
+var cass = require('cassandra-driver');
```
eevans: This filename is misspelled; `s/simeple.js/simple.js/`

test/mocha.opts
hardikj: Ah, I was wondering by tests are failing on travis, I was passing for me locally
reinstalling node_modules seems to give same behavior now.

eevans:
Hi @hardikj,

I'm still looking through the tests themselves, just to be sure we have parity with master, but I've added a few in-line comments to get you started.

In addition to the in-line comments, I had a few other initial observations:

All of the newly added test files are missing the special comments at the top of the file that serve to satisfy jshint that there are no missing requires.

    // mocha defines to avoid JSHint breakage
    /* global describe, it, before, beforeEach, after, afterEach */

Also, all of the newly add test files are missing the "use strict", at the top of the file.

Additionally, I'd like to see if we can work on some of these suite/test descriptions.  `describe()`, `context()`, and `it()` are all part of mocha's [BDD interface](http://mochajs.org/#bdd-interface).  The idea is that the string arg to `describe()` says *what* we are testing, a nested `context()` (functions identically to `describe()`) or `describe()` provides additional context, and the argument to `it()` describes the tested behavior behavior.  For example:

``` javascript
describe('Table operations', function() {
    context('Create', function() {
        it('successfully creates a simple test table', ...);
        it('throws an error for unsupported schemas', ...);
    });
});
```

It almost reads as english.  Create (which is a Table operation), successfully creates a simple table.  Create throws an error for unsupported schemas. Etc.

Thanks,

test/functional/schema_migrations.js
```diff
@@ -52,8 +52,8 @@ describe('Schema migration', function() {
         })
         .then(function(response) {
             assert.ok(response, 'undefined response');
+            console.log(response);
```
eevans: Remove this?

test/functional/schema_migrations.js
```diff
@@ -52,8 +52,8 @@ describe('Schema migration', function() {
...
             assert.deepEqual(response.status, 201);
-
             router.makeRouter();
```
eevans: Not your bug per say, but this is definitely wrong; Router should be initialized here the way it is in other tests.

hardikj:
@eevans, thanks for the review 

> All of the newly added test files are missing the special comments at the top of the file that serve to satisfy jshint that there are no missing requires.
>
>    // mocha defines to avoid JSHint breakage
>    /* global describe, it, before, beforeEach, after, afterEach */

I have left it on the top file [here](https://github.com/wikimedia/restbase-mod-table-cassandra/pull/110/files#diff-910eb6f57886ca16c136101fb1699231R5), thinking it will be recursive? 

> It almost reads as english. Create (which is a Table operation), successfully creates a simple table. >  Create throws an error for unsupported schemas. Etc

This makes sense and will improve the readability of tests a lot.

test/functional/secondaryIndexes.js
```diff
@@ -0,0 +1,358 @@
...
+                    method: 'get',
+                    body: {
+                        table: "simpleSecondaryIndexTable",
+                        limit: 1,
+                        next: next,
+                        attributes: {
+                            key: 'test2',
+                        }
+                    }
+                })
```
eevans: missing semi-colon

test/functional/secondaryIndexes.js
```diff
@@ -0,0 +1,358 @@
...
+                        next: next,
+                        attributes: {
+                            key: 'test2',
+                        }
+                    }
+                });
+            })
+            .then(function(response) {
+                deepEqual(response.body.items.length, 2);
+                var next = response.body.next
```
eevans: missing semi-colon

test/functional/simeple.js
```diff
@@ -0,0 +1,380 @@
...
+var TimeUuid = cass.types.TimeUuid;
+var db;
+
+describe('Db operations on a simple table', function() {
+
+    before(function () { 
+        return router.setup()
+        .then(function(newdb) {
+            db = newdb;
+        })
```
eevans: missing semi-colon

test/functional/simeple.js
```diff
@@ -0,0 +1,380 @@
...
+                    uri:'/restbase.cassandra.test.local/sys/table/simple-table/',
+                    method: 'get',
+                    body: {
+                        table: "simple-table",
+                        attributes: {
+                            key: 'test',
+                            tid: dbu.testTidFromDate(new Date('2013-08-08 18:43:58-0700'))
+                        }
+                    }
+                })
```
eevans: missing semi-colon

test/functional/types.js
```diff
@@ -0,0 +1,343 @@
...
+                response.body.items[0].boolean = true;
+            });
+        });
+        it('drop table', function() {
+            this.timeout(15000);
+            return router.request({
+                uri: "/restbase.cassandra.test.local/sys/table/typeTable",
+                method: "delete",
+                body: {}
+            })
```
eevans: missing semi-colon

PR #112, opened 2015-06-26
Normalise asynchronous updates and release v0.7.0

d00rman:
When a record is added/updated, the asynchronous updates kick in to update the secondary indices and enforce the declared retention policy. However, the original request's attributes are JS types, while all of the retrieved DB rows for which these actions are applied are in native, Cassandra representation, while the code supposed both were JS.

This PR fixes this issue by adding an additional parameter to `buildPutQuery()` and `buildCondition()` allowing users to supply their own conversion function which is applied to a request's query's attributes. This feature is used by the revision policy and the secondary index rebuilder, which supply `dbu.noConvert()`, a dummy conversion function simply returning the value passed in. If this extra parameter is not passed to `buildPutQuery()`, the default conversion scheme is used, thus not affecting the workflow of the other parts of the system.

Additionally, the request query attributes passed to the constructors of `RevisionPolicyManager` and `IndexRebuilder` are converted to the native Cassandra value types to avoid mixing Cassandra and JS types.

Also included in this PR:

- fix minor bug which allowed passing a string to `buildCondition()` instead of an Object - that was happening [here](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/517cd17b82ffdb0ec28d751f03305fbd5f61d104/lib/dbutils.js#L815)
- rename `_get()` to `_getRaw()` to make it clear that method does not perform attribute conversions
- remove the superfluous `async` module dependency
- release v0.7.0

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -627,18 +627,20 @@ dbu.convertRows = function convertRows (rows, schema) {
 };
 
 /**
- * Converts an internal request's query attributes to native
+ * Deep-clones and converts an internal request's query attributes to native
  * Cassandra representations
  *
  * @param {InternalRequest} internalReq the request whose attributes to convert
- * @param {Object} schema optionally, use this schema instead of internalReq.schema
- * @return {InternalRequest} the same request passed in, with converted values
+ * @return {InternalRequest} the clone of the request passed in, with converted values
  */
-dbu.convertWriteReqQueryAttrs = function(internalReq, schema) {
-    var conversions = (schema || internalReq.schema || {}).conversions;
-    var attrs = internalReq.query.attributes;
+dbu.makeConvertedRequest = function(internalReq) {
+    var conversions = (internalReq.schema || {}).conversions;
+    var clonedReq = internalReq.extend({
+        query: extend(true, {}, internalReq.query)
+    });
+    var attrs = clonedReq.query.attributes;
     if(!conversions || !attrs) {
-        return internalReq;
+        return clonedReq;
     }
     Object.keys(attrs).forEach(function(key) {
         var conv = conversions[key];
@@ -646,7 +648,7 @@ dbu.convertWriteReqQueryAttrs = function(internalReq, schema) {
             attrs[key] = conv.write(attrs[key]);
         }
     });
-    return internalReq;
+    return clonedReq;
 };
 
 /*
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -670,6 +670,14 @@ dbu.buildCondition = function buildCondition (predicates, schema, convertFunc) {
         }
     };
 
+    // make sure we have got a predicate object
+    if(!predicates || predicates.constructor !== Object) {
+        return {
+            cql: predicates || '',
+            params: []
+        };
+    }
+
     var params = [];
     var conjunctions = [];
     Object.keys(predicates).forEach(function(predKey) {
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -492,10 +492,6 @@ dbu.conversions = {
     uuid: { read: toString() }
 };
 
-dbu.noConvert = function(key, val) {
-    return val;
-};
-
 /*
  * Derive additional schema info from the public schema
  */
@@ -659,18 +655,18 @@ dbu.makeRawRequest = function(internalReq) {
  * CQL building for conditional requests in general.
  * @param {object} predicates, the 'attributes' object in queries.
  * @param {object} schema, the schema info for the logical table.
- * @param {Function} convertFunc the conversion function to use
+ * @param {Boolean} noConvert if true, no attribute value conversion will take place
  * @return {object} queryInfo object with cql and params attributes
  */
-dbu.buildCondition = function buildCondition (predicates, schema, convertFunc) {
-    convertFunc = convertFunc || function(key, val) {
+dbu.buildCondition = function buildCondition (predicates, schema, noConvert) {
+    function convert(key, val) {
         var convObj = schema.conversions[key];
-        if (convObj && convObj.write) {
+        if (!noConvert && convObj && convObj.write) {
             return convObj.write(val);
         } else {
             return val;
         }
-    };
+    }
 
     // make sure we have got a predicate object
     if(!predicates || predicates.constructor !== Object) {
@@ -692,7 +688,7 @@ dbu.buildCondition = function buildCondition (predicates, schema, convertFunc) {
         } else if (predObj === null || predObj.constructor !== Object) {
             // Default to equality
             cql += ' = ?';
-            params.push(convertFunc(predKey, predObj));
+            params.push(convert(predKey, predObj));
         } else {
             var predKeys = Object.keys(predObj);
             if (predKeys.length === 1) {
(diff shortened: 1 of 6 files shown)


d00rman:
I'm still in doubt if [lib/RevisionPolicy.js#L40](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/517cd17b82ffdb0ec28d751f03305fbd5f61d104/lib/revisionPolicy.js#L40) is correct or not, since, 'InternalRequest.extend()' does not deep-clone the `query` field, but on the next line its `query.attributes[key]` are being changed, hence modifying the original request's attributes. Should that be `extend(true, {}, this.request)` ?

d00rman:
cc @eevans @gwicke

lib/dbutils.js
```diff
@@ -630,16 +657,25 @@ dbu.convertRows = function convertRows (rows, schema) {
  * CQL building for conditional requests in general.
  * @param {object} predicates, the 'attributes' object in queries.
  * @param {object} schema, the schema info for the logical table.
+ * @param {Function} convertFunc the conversion function to use
```
gwicke: An alternative that's closer to dbu.conversions would be to pass in an object registry of conversions. By passing in an empty object, conversions can be disabled.

d00rman: That's true - we do not need (nor want, for that matter) to support *other* types of conversions. I'll modify that bit.

eevans:
> I'm still in doubt if lib/RevisionPolicy.js#L40 is correct or not, since, 'InternalRequest.extend()' does not deep-clone the query field, but on the next line its query.attributes[key] are being changed, hence modifying the original request's attributes. Should that be extend(true, {}, this.request) ?

I don't think the intention here was to get a copy of the request, but to simply add the `ttl` attribute.  The same couldn't be done with `query`, because `InternalRequest#extend` doesn't recurse.  In hindsight though, that does make for a confusing read.

lib/dbutils.js
```diff
@@ -622,6 +626,29 @@ dbu.convertRows = function convertRows (rows, schema) {
...
 
+/**
+ * Converts an internal request's query attributes to native
+ * Cassandra representations
+ *
+ * @param {InternalRequest} internalReq the request whose attributes to convert
+ * @param {Object} schema optionally, use this schema instead of internalReq.schema
+ * @return {InternalRequest} the same request passed in, with converted values
+ */
+dbu.convertWriteReqQueryAttrs = function(internalReq, schema) {
```
eevans: I wonder, would it be safer to return a deep copy of the internal request here, (rather than have the caller perform the copy)?  Is there a future use-case where we'd *want* to mutate the values of the original request?

eevans:
Overall this LGTM (see my one comment/question above though).

d00rman:
> I don't think the intention here was to get a copy of the request, but to simply add the ttl attribute. The same couldn't be done with query, because InternalRequest#extend doesn't recurse. In hindsight though, that does make for a confusing read

So why does the code then set `req.query.attributes[key] = row[key]`, with `row` being the evaluated row passed in to the method?

lib/dbutils.js
d00rman: I concur and I wanted to do it to begin with, but then ran into the *method naming* problem. How do I call such a method? `deepCopyAndConvert` (and the likes) seems a bit excessive. Given a good suggestion, I'm changing it right away :)

lib/revisionPolicy.js
```diff
@@ -12,7 +13,7 @@ var P = require('bluebird');
  */
 function RevisionPolicyManager(db, request, schema) {
     this.db = db;
-    this.request = request;
+    this.request = dbu.convertWriteReqQueryAttrs(extend(true, {}, request), schema);
```
gwicke: It might make sense to do this conversion once, before passing the converted request copy into RevisionPolicyManager *and* the IndexRebuilder.

We should also be able to drop the schema arg, as it's really the same as req.schema.

eevans:
> So why does the code then set req.query.attributes[key] = row[key], with row being the evaluated row passed in to the method?

I'm increasingly less sure that I know what you're asking about here.

lib/dbutils.js
eevans: Naming stuff is hard; I humbly propose the following (even though I like none of them):

* `getConvertedCopy`
* `convertedRequest`
* `convertedRequestCopy`
* `getConvertedRequest`
* `makeConvertedRequested`
* `convertedRequestFor`

In the end, as long as the semantics are well documented, I think we'd be OK.

d00rman:
> I'm increasingly less sure that I know what you're asking about here.

Hehe :) I'm talking about [this block of code](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/lib/revisionPolicy.js#L40-L43) where the original request's attributes are being overridden with the values from the current row being examined. Given what it does, and what you mentioned the idea was, I'm thinking something is awry. Or I'm baaadly misreading the code there.

lib/revisionPolicy.js
d00rman: > It might make sense to do this conversion once, before passing the converted request copy into RevisionPolicyManager and the IndexRebuilder.

What I'm concerned about here is that the revision policy handler mutates the original request object passed to its constructor, while the index rebuilder does not. Thus, if we create only one clone and distribute it to both handlers, we're probably going to have some edge-case bugs. That said, I am still not clear on whether the revision policy's mutation of the original request is intentional or not (cf. the discussion on this PR with @eevans ).

> We should also be able to drop the schema arg, as it's really the same as req.schema.

Yup, originally included it because the schema migration mechanism needs it (and the request is not an internal one) but neglected to remove it once I realised it doesn't fit the scheme of background jobs... Thnx for the observation!

d00rman:
@eevans @gwicke I moved the cloning into `dbu.makeConvertedRequest()` and optimised it so that only `req.query` is deep-cloned, for the rest `InternalReq.extend()` is used, which should minimise the memory footprint.

d00rman:
Finally settled on `makeRawRequest()` for the method name, and implemented @gwicke 's suggestion about passing in booleans instead of functions to `buildPutQuery()` and `buildCondition()`

d00rman:
@eevans I've tweaked the revision policy code a bit to address my own concerns (and also so you can see wth I was talking about :P) in [5cbd155](https://github.com/d00rman/restbase-mod-table-cassandra/commit/5cbd155d9a36a4711ea2e89d38cc817ef9fd7975). Take a look and see if that makes sense

lib/dbutils.js
```diff
@@ -630,18 +656,27 @@ dbu.convertRows = function convertRows (rows, schema) {
...
             return convObj.write(val);
         } else {
             return val;
         }
     }
 
+    // make sure we have got a predicate object
+    if(!predicates || predicates.constructor !== Object) {
+        return {
+            cql: predicates || '',
```
gwicke: I'm not sure if we should support passing in a raw query here. I fear that it would cover up bugs, without solving an actual problem.

d00rman: This may already be happening. See the beginning and the end of [this block](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/lib/dbutils.js#L793-L819): first, `query.if` is explicitly allowed to be a string, but then later on, we seem to assume it's an object since we use `buildCondition(query.if, schema)`

eevans:
LGTM :+1:

lib/dbutils.js
gwicke: I think that is actually a bug. The string case should only pertain to 'if not exists', but not to the query part where we perform a CAS by comparing with existing values, specified in an object. We should throw an error if query.if is a string, but does not normalize to 'if not exists'.

d00rman: k, will add that

d00rman:
@gwicke fixed the issue, errors are now thrown if the string is not equal to //not exists// or if something other than an Object is provided.

@eevans heads up wrt your test suite PR / ongoing work, I added two new tests in b608e36

gwicke:
LGTM, :+1: from me. Waiting for Travis with the merge, but am keen to get this onto staging soon so that we can push it out to prod today in case that works fine.

gwicke:
Travis is quite slow, and I tested this locally. So, going ahead with the merge.

PR #113, opened 2015-06-29
Bound memory used by background updates

gwicke:
- Set a small fetch size of 5 rows at a time
- Break up the promise chain at each page boundary


eevans:
LGTM

PR #115, opened 2015-06-30
Bump version after re-upload to fix sha1sum

gwicke:
Bug: https://phabricator.wikimedia.org/T104284


gwicke:
Trivial, so self-merging.

PR #116, opened 2015-06-30
remove old revisions

eevans:


maintenance/thin_out_key_rev_value_data.js
```diff
@@ -100,6 +100,8 @@ function processRow (row) {
     }
     keys = newKeys;
 
+    console.log('processing: ', keys.rev, new Date(row.tid.getDate()));
```
gwicke: This print will slow down the script a lot.

PR #117, opened 2015-07-03
Make exception handling more robust for background updates and release v0.7.3

d00rman:
Bug: [T104581](https://phabricator.wikimedia.org/T104581)

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -496,7 +496,7 @@ DB.prototype._put = function(req) {
         if (schema.secondaryIndexes) {
             self._backgroundUpdates(req, 3)
             .catch(function(err) {
-                self.log('error/cassandra/rebuildIndexes', err);
+                self.log('error/cassandra/backgroundUpdates', err);
             });
         }
 
@@ -592,7 +592,7 @@ DB.prototype._backgroundUpdates = function(req, limit, indexes) {
             return P.each(res.items.reverse(), handler.handleRow.bind(handler));
         }).catch(function(err) {
             // just log it
-            self.log('error/cassandra/backgroundupdates', err);
+            self.log('error/cassandra/backgroundUpdates', err);
         });
     })
     .then(function() {       // Query for 'limit' subsequent records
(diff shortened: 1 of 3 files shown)


lib/db.js
```diff
@@ -588,7 +588,11 @@ DB.prototype._backgroundUpdates = function(req, limit, indexes) {
     // matching row, in descending order.
     return self._getRaw(newerRebuildRequest)
     .then(function(res) {    // Query for one record previous
-        return P.each(res.items.reverse(), handler.handleRow.bind(handler));
+        return P.try(function() {
+            return P.each(res.items.reverse(), handler.handleRow.bind(handler));
+        }).catch(function() {
+            // noop here for now
```
Pchelolo: Could add logging here, at least we would know smth happened.

Pchelolo:
LGTM, except one suggestion about logging.

lib/db.js
d00rman: Done

d00rman:
I have released v0.7.2 in order to be able to re-deploy the last version of RESTBase with only the new version of the storage module.

d00rman:
v0.7.3 has been released to npm, https://gerrit.wikimedia.org/r/223297 is the deployment candidate only bumping the version

lib/dbutils.js
```diff
@@ -122,14 +122,16 @@ dbu.makeSchemaHash = function makeSchemaHash(schema) {
...
+        return client.execute_p(query, params, {
+            prepare: true,
+            fetchSize: options.fetchSize || 5,
+            pageState: pageState,
+        });
     })
     .catch(function(err) {
         if (!options.retries) {
-            throw err;
+            return P.reject(err);
```
gwicke: Afaik this is equivalent to throw, as it is inside a promise handler.

lib/dbutils.js
```diff
@@ -150,19 +152,29 @@ function _nextPage(client, query, params, pageState, options) {
...
         return _nextPage(client, query, params, pageState, options)
         .then(function(res) {
-            return P.each(res.rows, handler)
-            .then(function() {
-                if (res.pageState === null) {
+            return P.try(function() {
+                return P.each(res.rows, handler);
+            }).catch(function(e) {
+                // something fishy's going on, abort abort abort!
+                return P.reject(e);
```
gwicke: Same here; you are basically re-raising the error.

gwicke:
LGTM overall, just two small cases where I think a .catch is not needed / doesn't make a difference.

d00rman:
@gwicke forgot to remove these useless (and desperate) changes. Removed them.

PR #118, opened 2015-07-08
More tests refactoring in preparation to move to a separate module

Pchelolo:
More refactoring is needed on tests to allow moving them to a separate module.

- Replaced cassandra driver's uuid with our cassandra-uuid lib. As constructor is checked by driver, replaced TimeUuid objects with strings everywhere as strings should be correctly handled by both implementations. 
- Removed dependency from functional tests to dbutils: moved testTidFromDate to test_utils as it's used only by tests. Also removed the use of hash function and started to compare plain schema objects which should be fine bc 'hash match' === 'objects match'
- 2 tests were failing - they received unexpected `revisionRetentionPolicy = [ undefined ]`, so I set this schema property only when it's non-empty.
- Fixed one before which didn't return the promise and so could swallow an error.
- Refactored all functional tests to drop dependency from DB and replaced it with calls to router
- Moved one tests from functional to unit, because it depends on deleting a row for table which doesn't yet have a counterpart in routes.

I'm creating this PR to avoid review hell if I change & move files to the separate module simultaneously.

After this functional tests depend only on the test_router, which depends on the library constructor. Constructor would be passed to testing module as parameter, so after this refactoring tests are ready to be moved to separate module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ npm-debug.log
 .tern-project
 .tern-port
 *~
+.idea/*
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -2,10 +2,6 @@
 var crypto = require('crypto');
 var extend = require('extend');
 var cass = require('cassandra-driver');
-var Uuid = cass.types.Uuid;
-var TimeUuid = cass.types.TimeUuid;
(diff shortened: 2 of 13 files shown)


Pchelolo:
cc @gwicke @eevans @d00rman

PR #119, opened 2015-07-09
Moved tests to a separate module

Pchelolo:
The tests and the spec are moved to a separate module:
https://github.com/wikimedia/restbase-mod-table-spec

Along with this review, please have a look at the repository with tests&spec.

The only problem is that now scripts/direct-request doesn't work, but it seems
broken and outdated, so I've decided to wait on fixing it until I know more details about it.


package.json
```diff
@@ -18,14 +18,12 @@
...
-    "core-js": "^0.5.4",
+  "devDependencies": {
     "coveralls": "2.11.2",
-    "routeswitch": "~0.6.3",
     "istanbul": "0.3.5",
     "mocha-lcov-reporter": "0.0.1",
     "mocha": "x.x.x",
     "mocha-jshint": "0.0.9",
-    "cassandra-uuid":"^0.0.2"
+    "cassandra-mod-table-spec": "git+https://github.com/wikimedia/restbase-mod-table-spec#master"
```
d00rman: The module is used in `index.js`, so it needs to go into `dependencies`

PR #120, opened 2015-07-10
Hotfix: Force limit to actually work on secondary index updates

gwicke:
The missing limit functionality here is causing rewrite bursts with on the
order of hundreds of queries per second, which causes OOM and compaction to
lag behind.

This is a hotfix, intended to stop the bleeding until we can clean this up
more thoroughly.

Bug: https://phabricator.wikimedia.org/T105509


eevans:
+1 as a temporary hotfix

PR #122, opened 2015-07-14
avoid unnecessary TTL rewrites

eevans:
Refactored dbu#buildGetQuery to treat the default projection (*), as an
explicit set of attributes.  This avoids the special-casing for 2.0.x versions,
and makes appending TTL attributes more straightforward.

Introduced a boolean parameter to db#_getRaw and dbu#buildGetQuery to specify
results that include a TTL attribute for all candidates (candidates being any
non-index, non-collection attribute).

Added TTL results to the queries performed in db#_backgroundUpdates, and
utilized them to avoid rewriting rows that have already had a TTL set.

Bug: T105509

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -997,6 +997,14 @@ dbu.buildGetQuery = function(req, options) {
         }
     }
 
+    // Generally, req.query.limit is used to limit per-page results, which is
+    // managed through the driver's pageState.  When it's necessary to use a
+    // CQL LIMIT, it should be included in options.
+    if (options && options.limit) {
+        var limit = parseInt(options.limit);
+        cql += limit ? ' limit ' + limit : '';
+    }
+
     return {cql: cql, params: params};
 };
 
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -901,15 +901,15 @@ dbu.buildGetQuery = function(req, options) {
         req.columnfamily = dbu.idxColumnFamily(query.index);
     }
 
-    var projStr = Object.keys(schema.attributes).map(dbu.cassID).join(',');
+    var projCQL = Object.keys(schema.attributes).map(dbu.cassID).join(',');
     var projAttrs = schema.attributes;
 
     if (query.proj) {
         if (Array.isArray(query.proj)) {
-            projStr = query.proj.map(dbu.cassID).join(',');
+            projCQL = query.proj.map(dbu.cassID).join(',');
             projAttrs = query.proj;
         } else if (query.proj.constructor === String) {
-            projStr = dbu.cassID(query.proj);
+            projCQL = dbu.cassID(query.proj);
             projAttrs = [query.proj];
         }
     }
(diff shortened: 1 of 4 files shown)


lib/dbutils.js
```diff
@@ -878,16 +882,16 @@ dbu.buildPutQuery = function(req, noConvert) {
 /**
  * CQL building for GET queries
  * @param {InternalRequest} req
+ * @param {Boolean} withTTL if true, return TTLs of non-index attributes
  * @return {object} queryInfo object with cql and params attributes
  */
-dbu.buildGetQuery = function(req) {
-    var proj = '*';
-
+dbu.buildGetQuery = function(req, withTTL) {
```
gwicke: I'm wondering if an options object would be clearer, especially when calling buildGetQuery.

lib/dbutils.js
```diff
@@ -897,26 +901,40 @@ dbu.buildGetQuery = function(req) {
...
+            projAttrs = [query.proj];
         }
     }
 
+    // Add TTL attributes for all non-index attributes
+    if (options && options.withTTLs) {
+        // Candidates for TTL are non-index, non-collection, attributes
+        var ttlCandidates = Object.keys(projAttrs).filter(
+            function(v) {
+                return !schema.iKeyMap[v] && !/(set|map|list)<.+>/.test(schema.attributes[v]);
```
gwicke: `/^(set|map|list)<.*>$/`

lib/dbutils.js
```diff
@@ -897,26 +901,40 @@ dbu.buildGetQuery = function(req) {
         req.columnfamily = dbu.idxColumnFamily(query.index);
     }
 
+    var projStr = Object.keys(schema.attributes).map(dbu.cassID).join(',');
```
gwicke: `projCQL` might be slightly clearer.

lib/revisionPolicy.js
```diff
@@ -45,9 +45,21 @@ RevisionPolicyManager.prototype.handleRow = function(row) {
     // creates a deep clone of self.request.query, so we are sure not to
     // modify the original incoming request
     var attrs = self.request.query.attributes;
+    var refCount = 0;
     Object.keys(attrs).forEach(function(key) {
+        var id = dbu.cassTTL(key);
```
gwicke: Should this live in _getRaw?

lib/revisionPolicy.js
```diff
@@ -45,9 +45,21 @@ RevisionPolicyManager.prototype.handleRow = function(row) {
...
+        if (row.hasOwnProperty(id) && (!row[id] || row[id] >= self.policy.grace_ttl)) {
+            refCount++;
+        }
         attrs[key] = row[key];
     });
+
+    // If there were no unset TTLs, or TTLs that would have been lowered
+    // according to the current policy, then avoid a redundant update and bail.
```
gwicke: This doesn't look correct if the row contains collections.

PR #123, opened 2015-07-15
Compare with a fixed date

gwicke:
When thinning both html and data-parsoid it's easier to maintain a clean
cut-off when comparing with a fixed date.


eevans:
Yeah, good point (see also: https://phabricator.wikimedia.org/T105737)

PR #127, opened 2015-07-17
Implement exponential retry back-off

gwicke:
- add exponential back-off, so that temporary issues are more likely to
  resolve themselves
- match >= on tokens and domain/key matches for more effective retry on an
  existing key after a failure


eevans:
LGTM

PR #132, opened 2015-07-24
Improve the thin-out script

gwicke:
- automatically step over problematic token ranges
- consistently report the current token range

### DIFF
--- a/maintenance/thin_out_key_rev_value_data.js
+++ b/maintenance/thin_out_key_rev_value_data.js
@@ -84,6 +84,26 @@ var keys = {
 
 var total = 0;
 
+// Parse optional start offsets
+var startOffset = {
+    token: null,
+    domain: null,
+    key: null,
+    pageState: null,
+};
+
+if (/^-?[0-9]{1,30}$/.test(process.argv[4])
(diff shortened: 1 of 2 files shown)


maintenance/thin_out_key_rev_value_data.js
```diff
@@ -183,12 +194,21 @@ function nextPage(pageState, retryDelay) {
     })
     .catch(function(err) {
         retryDelay = retryDelay || 1; // ms
-        // Retry at least roughly once per 30 minutes
-        if (retryDelay < 900 * 1000) {
+        if (retryDelay < 20 * 1000) {
```
d00rman: Put 20000 right away ?

d00rman:
LGTM, but jshint complains *'startOffset' was used before it was defined. at maintenance/thin_out_key_rev_value_data.js:149* so its declaration should be moved up before `processRow()`

maintenance/thin_out_key_rev_value_data.js
gwicke: Maybe.. it's harder to read when the # of seconds is larger. Should be optimized away anyway.

PR #133, opened 2015-07-25
Make background updates synchronous

gwicke:
With idempotent updates it is better to retry the put at a higher level if
necessary. Consumers can still choose to ignore failures, but now have a
chance to repeat requests in case any part of the request processing fails.

This change should provide better back-pressure in write overload situations
by slowing those requests down until the request is fully processed. As a
consequence, it should reduce the memory build-up in restbase when writes are
slow.

As a 'small' bug fix, this patch also makes sure that background updates are run
whenever a retention policy other than 'all' is specified, even if no
secondary indexes are defined on the table.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -479,11 +479,11 @@ DB.prototype._put = function(req) {
 
     //console.log(batch, schema);
     var queryOptions = {consistency: req.consistency, prepare: true};
-    var mainUpdate;
+    var update;
     if (batch.length === 1) {
         // Single query only (no secondary indexes): no need for a batch.
         var queryInfo = batch[0];
-        mainUpdate = this.client.execute_p(queryInfo.cql, queryInfo.params, queryOptions);
+        update = this.client.execute_p(queryInfo.cql, queryInfo.params, queryOptions);
     } else {
         var driverBatch = batch.map(function(queryInfo) {
             return {
@@ -491,26 +491,40 @@ DB.prototype._put = function(req) {
                 params: queryInfo.params
             };
         });
-        mainUpdate = this.client.batch_p(driverBatch, queryOptions);
+        update = this.client.batch_p(driverBatch, queryOptions);
     }
 
-    return mainUpdate
+    if (self._shouldRunBackgroundUpdates(schema))
+    {
+        update = update.then(function() {
+            return self._backgroundUpdates(req, 3);
+        });
+    }
 
-    .then(function(result) {
-        // Kick off asynchronous local index rebuild
-        if (schema.secondaryIndexes) {
-            self._backgroundUpdates(req, 3)
-            .catch(function(err) {
-                self.log('error/cassandra/backgroundUpdates', err);
-            });
-        }
-
-        // But don't wait for it. Return success straight away.
+    return update.then(function(result) {
         return {
             // XXX: check if condition failed!
             status: 201
         };
     });
+};
+
+/*
+ * Predicate: should we run background updates?
+ *
+ * @param {object} schema; the table schema
+ * @param {array} (optional) indexes to process
+ * @return {boolean} whether to apply background updates
+ */
+DB.prototype._shouldRunBackgroundUpdates = function(schema, indexes) {
+    if (!indexes) {
+        indexes = Object.keys(schema.secondaryIndexes);
+    }
+    // If there are no indexes, and the retention policy is 'all' (i.e.
+    // there are no revisions to cull), then there is no need to run
+    // background updates.
+    return indexes.length
+        || schema.revisionRetentionPolicy && schema.revisionRetentionPolicy.type !== 'all';
 };
 
 /*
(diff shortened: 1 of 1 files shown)


d00rman:
I both agree and disagree with this PR. It does indeed help in overload situations as the caller needs to wait for their completion. What bothers me here is that these background updates are now a mandatory part of the insert/update, whereas I think they should be on a best-effort basis. It is not essential that they succeed, given that storing a new re-render for the same revision will *correct* any previous attempts that might have failed. Moreover, I don't think it is correct to propagate any failures from background updates to the client if the main update succeeded, as this conveys the wrong information to the client; only main update failures should be propagated.

eevans:
@d00rman 
> I both agree and disagree with this PR. It does indeed help in overload situations as the caller needs to wait for their completion. What bothers me here is that these background updates are now a mandatory part of the insert/update, whereas I think they should be on a best-effort basis. It is not essential that they succeed, given that storing a new re-render for the same revision will correct any previous attempts that might have failed. Moreover, I don't think it is correct to propagate any failures from background updates to the client if the main update succeeded, as this conveys the wrong information to the client; only main update failures should be propagated.

I understand (and sympathize) with the entirety of this, but am still in favor of this being merged.  Each write has a cost, and we need the back-pressure that in-lining that cost will provide.

@gwicke 
As for the changeset itself, a test of the new behavior would be good; A test that PUTs and validates the expected exception.

gwicke:
@d00rman, a new re-render might take a long time (months), during which time the secondary index will be incorrect. I agree that we should handle retries internally once we have the capability to do so cleanly (-> event queue), but right now we don't have a way to reliably record the need to retry later. 

However, I'm not convinced that the table storage layer is the right layer for those retries and/or error propagation decisions. Whether it is safe to ignore secondary index update failures or not depends on the use case. For example, in the case of Parsoid HTML, we need to be sure that HTML and data-parsoid are stored safely & are accessible through the indexes we need, so that saves from VE based on that HTML does not result in corruptions.

gwicke:
@eevans:
> A test that PUTs and validates the expected exception.

You mean a PUT that triggers a storage backend failure or timeout? I guess we could try to inject faults, but I'm a bit wary of the ugliness that might add to the code. Do you see a clean way to do fault injection that doesn't involve hacking up db.js?

eevans:
> Do you see a clean way to do fault injection that doesn't involve hacking up db.js?

No, I suppose not; The best I can think of would be to use dependency injection to mock `RevisionPolicyManager` or `IndexRebuilder`, but that looks like the sort of slippery slope sure to significantly alter the scope of this PR, so disregard.

----

I'm :+1: on this request.

gwicke:
@d00rman, are you okay with this as-is, or would you prefer to return HTML even if the save wasn't fully successful?

d00rman:
> @d00rman, are you okay with this as-is, or would you prefer to return success from the table storage backend even if the save wasn't fully successful?

I'm ok with making the updates synchronous, but would not like to see the failures propagated to the caller if the main update succeeded. Some options might be:

- return the code as reported by the main update
- return custom codes in case the bkg update fails so that the client can handle that specific case if needed

gwicke:
How about letting the request indicate which level of reliability / consistency it needs, and then returning success / error based on that? There is already some embryonic support for that with the `consistency` query property in the table storage backend, but so far that's only applied to the C* queries and is not fully specified or robust. As far as I know it is also not used for anything in RESTBase.

I guess the main use cases are 'strongly consistent', 'eventually consistent', plus (perhaps?) a 'fire & forget' level.

d00rman:
> How about letting the request indicate which level of reliability / consistency it needs, and then returning success / error based on that?

Great idea.

> There is already some embryonic support for that with the consistency query property in the table storage backend, but so far that's only applied to the C* queries and is not fully specified or robust. As far as I know it is also not used for anything in RESTBase. I guess the main use cases are 'strongly consistent', 'eventually consistent', plus (perhaps?) a 'fire & forget' level.

I wouldn't use `consistency` though, because we use it in the storage back-end with a meaning not compatible with what we are after here. Both address consistency, but with slightly different targets. How about using `dataConsistency` ? With the values you suggest:

- `strong` - do the updates synchronously and return `2xx` *iff* both the main and background updates succeed, i.e. return the result returned by the background update
- `eventual` - do the updates synchronously and return the result of the main update
- `loose` (or `quick` or even `none`) - do the main update synchronously and the background update asynchronously and return the result of the main update

Makes sense?

gwicke:
@d00rman, are you on board with defaulting to strongly consistent / synchronous updates?

gwicke:
As an additional data point, DynamoDB defaults to eventual consistency for writes. However, their definition is that it's going to happen soon, without further writes.

@Pchelolo @d00rman @eevans, how about starting with strong consistency until we have real eventual consistency with queuing, and then switching the default?

d00rman:
I'd prefer to see using eventual consistency for tables that don't have secondary indices attached to them (as rev policy can hardly be seen as a factor for consistency), but I'm aware of the fact that it kind of violates the KISS principle, so the KISSer in me says *Okay*

gwicke:
Self-merged per @d00rman and @eevans's +1s

d00rman:
This PR seems to have broken RESTBase travis tests (or rather, releasing #134). Interestingly, @Pchelolo and I cannot reproduce the time-out error locally.

gwicke:
@d00rman, #292 disables timeouts in our mocha tests.

PR #135, opened 2015-08-01
Various fixes in background updates

gwicke:


lib/db.js
```diff
@@ -604,15 +613,21 @@ DB.prototype._backgroundUpdates = function(req, limit, indexes) {
     // Query for a window that includes 1 newer record (if any exists), and up
     // to 'limit' later records.  Run the list of update handlers across each
     // matching row, in descending order.
-    return self._getRaw(newerRebuildRequest, { withTTLs: true })
+    return self._getRaw(newerRebuildRequest)
     .then(function(res) {    // Query for one record previous
         return P.try(function() {
-            return P.each(res.items.reverse(), handler.handleRow.bind(handler));
+            return P.each(res.items.reverse(), indexRebuilder.handleRow.bind(indexRebuilder));
         }).catch(function(err) {
```
Pchelolo: We could probably merge this catch with a one on line 625. IndexRebuilder shouldn't throw 404, so the behaviour shouldn't change

gwicke: Done in the amend.

PR #136, opened 2015-08-07
Use common validator

Pchelolo:
Use a common validator for schema validation. Also, adds support for where conditions like in [docs](https://github.com/wikimedia/restbase/blob/master/doc/TableStorageAPI.md)
```
{
    table: 'example',
    attributes: {
        name: 'Tom',
        property: {
            gt: 'a',
            le: 'c'
        }
    },
    // Only select the 'value' column
    proj: ['value']
    limit: 50
}
```
(note two conditions in `property`) 

This PR depends on a spec PR, so travis is expected to fail.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -9,6 +9,7 @@ var cassID = dbu.cassID;
 var revPolicy = require('./revisionPolicy');
 var SchemaMigrator = require('./schemaMigration');
 var secIndexes = require('./secondaryIndexes');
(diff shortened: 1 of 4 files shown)


d00rman:
LGTM :+1:

@eevans @gwicke good for you too?

eevans:
Yeah, LGTM

d00rman:
Published on npmjs as v0.7.9

PR #137, opened 2015-08-17
Implemented 'interval' policy support

Pchelolo:
The PR implements an interval option for revision retention policy for Cassandra backend using and algorithm, described in a ticket.

Note: this PR depends on spec one: wikimedia/restbase-mod-table-spec#12

Bug: https://phabricator.wikimedia.org/T94524

### DIFF
--- a/lib/revisionPolicy.js
+++ b/lib/revisionPolicy.js
@@ -19,8 +19,13 @@ function RevisionPolicyManager(db, request, schema) {
     // The latest / just inserted row is not processed by the retentionPolicy,
     // so start with count 1.
     this.count = 1;
-    this.candidates = [];
-    this.candidatesFlushed = false;
+    if (this.policy.type === 'interval') {
+        var interval = this.policy.interval * 1000;
+        var tidTime = this.request.query.timestamp;
+        this.intervalLimitTime = tidTime - tidTime % interval;
+    } else {
+        this.intervalLimitTime = null;
+    }
 }
 
 /**
@@ -67,15 +72,10 @@ RevisionPolicyManager.prototype.handleRow = function(row) {
         }
         return setTtl(row);
     } else if (this.policy.type === 'interval') {
-        var interval = this.policy.interval * 1000;
-
-        var tidTime = this.request.query.timestamp;
-        var intervalLimitTime = new Date(tidTime - interval);
-        if (row[this.schema.tid].getDate() > intervalLimitTime && !row._ttl) {
-            this.candidates.push(row);
-        } else if (!this.candidatesFlushed) {
-            this.candidatesFlushed = true;
-            return P.all(this.candidates.map(setTtl));
+        if (row[this.schema.tid].getDate() > this.intervalLimitTime && !row._ttl) {
+            return setTtl(row);
+        } else {
+            return P.resolve();
         }
     }
 };
--- a/lib/revisionPolicy.js
+++ b/lib/revisionPolicy.js
@@ -19,6 +19,8 @@ function RevisionPolicyManager(db, request, schema) {
     // The latest / just inserted row is not processed by the retentionPolicy,
     // so start with count 1.
     this.count = 1;
+    this.candidates = [];
+    this.candidatesFlushed = false;
 }
 
 /**
@@ -30,6 +32,19 @@ function RevisionPolicyManager(db, request, schema) {
 RevisionPolicyManager.prototype.handleRow = function(row) {
     var self = this;
 
+    function setTtl(item) {
+        self.request.query.attributes = item;
+        self.request.query.timestamp = null;
+
+        var query = dbu.buildPutQuery(self.request, true);
+        var queryOptions = {consistency: self.request.consistency, prepare: true};
+
+        return self.db.client.execute_p(query.cql, query.params, queryOptions)
+        .catch(function(e) {
+            self.db.log('error/table/cassandra/revisionRetentionPolicyUpdate', e);
+        });
+    }
+
     if (self.noop) {
         return P.resolve();
     }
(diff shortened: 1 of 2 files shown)


lib/revisionPolicy.js
```diff
@@ -39,26 +54,30 @@ RevisionPolicyManager.prototype.handleRow = function(row) {
...
-    .catch(function(e) {
-        self.db.log('error/table/cassandra/revisionRetentionPolicyUpdate', e);
-    });
+        var tidTime = this.request.query.timestamp;
+        var intervalLimitTime = new Date(tidTime - interval);
+        if (row[this.schema.tid].getDate() > intervalLimitTime && !row._ttl) {
+            this.candidates.push(row);
+        } else if (!this.candidatesFlushed) {
+            this.candidatesFlushed = true;
+            this.candidates.forEach(setTtl);
```
eevans: Hrmm, this algorithm does seem right; Perhaps I don't understand how it's intended to work.  How does `count` factor in here?  Isn't it the objective to keep `count` records for each `interval` time period?

Also, since you're not returning the promise(s) here, this is completely async from the caller.

Pchelolo: @eevans The count is handled [earlier] (https://github.com/wikimedia/restbase-mod-table-cassandra/pull/137/files#diff-965579902a6fa2635a4e35061074f682R52), the same way for for latest & interval , so we get here only for renders which are `count + 1` etc. 

As for not returning, you are right for sure, will fix in a moment

Pchelolo:
@eevans updated, now we return the result

lib/revisionPolicy.js
eevans: > @eevans The count is handled earlier, the same way for for latest & interval , so we get here only for renders which are count + 1 etc. 

Ok, so this is like `latest`, but for a bounded time period?

lib/revisionPolicy.js
```diff
@@ -39,26 +54,30 @@ RevisionPolicyManager.prototype.handleRow = function(row) {
...
-    .catch(function(e) {
-        self.db.log('error/table/cassandra/revisionRetentionPolicyUpdate', e);
-    });
+        var tidTime = this.request.query.timestamp;
+        var intervalLimitTime = new Date(tidTime - interval);
+        if (row[this.schema.tid].getDate() > intervalLimitTime && !row._ttl) {
+            this.candidates.push(row);
+        } else if (!this.candidatesFlushed) {
+            this.candidatesFlushed = true;
+            return P.all(this.candidates.map(setTtl));
```
gwicke: This will update all candidates at once; I think it might be better to limit the concurrency. One option would be doing one at a time with Promise.each, the other would be Promise.map with a concurrency limit set (see https://github.com/petkaantonov/bluebird/blob/master/API.md).

lib/revisionPolicy.js
```diff
@@ -30,6 +32,19 @@ function RevisionPolicyManager(db, request, schema) {
 RevisionPolicyManager.prototype.handleRow = function(row) {
     var self = this;
 
+    function setTtl(item) {
```
gwicke: This doesn't seem to close over anything. It might be clearer to move it to a regular method instead.

lib/revisionPolicy.js
eevans: > @eevans The count is handled earlier, the same way for for latest & interval , so we get here only for renders which are count + 1 etc. 

Ok, so this is like `latest`, but for a bounded time period?

I'm curious:  Is there a scenario where `(row[this.schema.tid].getDate() > intervalLimitTime && !row._ttl)` would be true for every invocation of `handleRow`, if so, how will the flush be triggered?  And, considering that the window of records that a policy hander is ran against is at most 5, what would happen for a policy where `count` was say 7?

lib/revisionPolicy.js
gwicke: I don't think a count other than 1 makes sense for the interval policy.


>  Is there a scenario where (row[this.schema.tid].getDate() > intervalLimitTime && !row._ttl) would be true for every invocation of handleRow

Yes, definitely. In the current setup, that would lead to nothing being deleted. I think it would make more sense to delete everything but the last entry. However, to figure out whether it really is the last entry we would need to change the background update interface so that the background updates can control further iterative querying / paging. Maybe we can return a specific value like some CONTINUE constant that will keep the outer iterative query going?

gwicke:
>  Is there a scenario where (row[this.schema.tid].getDate() > intervalLimitTime && !row._ttl) would be true for every invocation of handleRow

Yes, definitely. In the current setup, that would lead to nothing being deleted. I think it would make more sense to delete everything but the last entry. However, to figure out whether it really is the last entry we would need to change the background update interface so that the background updates can control further iterative querying / paging. Maybe we can return a specific value like some CONTINUE constant that will keep the outer iterative query going?

> And, considering that the window of records that a policy hander is ran against is at most 5, what would happen for a policy where count was say 7?

I don't think a count other than 1 makes sense for the interval policy.

lib/revisionPolicy.js
```diff
@@ -39,26 +54,30 @@ RevisionPolicyManager.prototype.handleRow = function(row) {
...
-    var query = dbu.buildPutQuery(self.request, true);
-    var queryOptions = { consistency: self.request.consistency, prepare: true };
-
-    return self.db.client.execute_p(query.cql, query.params, queryOptions)
-    .catch(function(e) {
-        self.db.log('error/table/cassandra/revisionRetentionPolicyUpdate', e);
-    });
+        var tidTime = this.request.query.timestamp;
+        var intervalLimitTime = new Date(tidTime - interval);
+        if (row[this.schema.tid].getDate() > intervalLimitTime && !row._ttl) {
```
gwicke: Actually, I think we need to keep going if we see a row with a TTL within the time limit, to figure out the timestamp of the next non-deleted row. Based on that timestamp, we then need to revisit the decision on what to keep, to make sure that we do indeed keep one copy per 24 hours.

lib/revisionPolicy.js
Pchelolo: @eevans Actually, this is a part of the algo.

The original idea is that it's hard to 'start' the algo, so if we, say, go over all renders within `interval` from now and keep only `number` of them, we have the following problem: say interval is 10 seconds and `number` is 1. At second 1 comes a render, then at second 2 comes a render - we remove the first, now our `t0` is second 2 and the beginning shifts, at second 3 everything happens again and t0 shifts more and more - we can't start.

The problem that I've just realised if we have 10000 renders on interval 1, than interval2 has no renders, and interval3 has renders again - 10000 renders on day 1 would never get removed, so this indeed needs to be revisited.

lib/revisionPolicy.js
```diff
@@ -39,26 +60,25 @@ RevisionPolicyManager.prototype.handleRow = function(row) {
...
+        // the requests are sequentially sent, and once executed, the request
+        // object itself is not needed. Also, note that dbu.makeRawRequest()
+        // creates a deep clone of self.request.query, so we are sure not to
+        // modify the original incoming request
+        if (row._ttl && row._ttl <= this.policy.grace_ttl) {
+            return P.resolve();
+        }
+        return self._setTtl(row);
+    } else if (this.policy.type === 'interval') {
+        if (row[this.schema.tid].getDate() > this.intervalLimitTime && !row._ttl) {
```
d00rman: `s/>/>=/` because intervals are actually `[0, interval - 1]`, `[interval, 2 * interval - 1]`, etc...

PR #138, opened 2015-08-17
multi-datacenter replication

eevans:
In `DB#createTable`, validate that current keyspace replication meets requirements (based on the list of datacenters from module configuration, and any durability options in the request), and adjust if necessary.

Bug: https://phabricator.wikimedia.org/T108613

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -687,6 +687,107 @@ DB.prototype._migrateBackend = function(req, from, to) {
     }
 };
 
+/**
+ * Conditionally performs a table schema and/or back-end migration.
+ *
+ * @param  {object} req;               the current request object
+ * @param  {object} currentSchemaInfo; schema info object for current schema
+ * @param  {object} newSchema;         the proposed schema
+ * @param  {object} newSchemaInfo;     schema info object for proposed schema
+ * @return {object} HTTP response
+ */
+DB.prototype._migrateIfNecessary = function(req, currentSchemaInfo, newSchema, newSchemaInfo) {
+    var self = this;
+
+    var migrationPromise = P.resolve();
+    var emptyPromise = migrationPromise;
+
+    // The _backend_version attribute is excluded from the schema hash
+    // calculation so that we can evaluate these different classes of
+    // change (_backend_version versus table schema) separately.  If
+    // backend versions differ, a backend migration will need to be
+    // performed first.  Afterward, if and only if there has been a
+    // change to the schema (that is if the hashes do not match), will
+    // a schema migration occur. These schema changes must also include
+    // a monotonically increasing version number.  Finally, if either
+    // (or both) of a backend or table schema migration occur, the new
+    // JSON blob will be persisted.
+
+    // First carry out any back-end migration (if needed).
+    if (currentSchemaInfo._backend_version !== newSchemaInfo._backend_version) {
+        // A downgrade (unsupported)!
+        if (newSchemaInfo._backend_version < currentSchemaInfo._backend_version) {
+            throw new dbu.HTTPError({
+                status: 400,
+                body: {
+                    type: 'bad_request',
+                    title: 'Unable to downgrade storage backend to version '+newSchemaInfo._backend_version,
+                    keyspace: req.keyspace,
+                    schema: newSchema
+                }
+            });
+        }
+        
+        migrationPromise = self._migrateBackend(req, currentSchemaInfo, newSchemaInfo);
+    }
+
+    // Next carry out any table schema migration (if needed).
+    if (currentSchemaInfo.hash !== newSchemaInfo.hash) {
+        var migrator;
+        try {
+            migrator = new SchemaMigrator(
+                self,
+                req,
+                currentSchemaInfo,
+                newSchemaInfo);
+        }
+        catch (error) {
+            throw new dbu.HTTPError({
+                status: 400,
+                body: {
+                    type: 'bad_request',
+                    title: 'The table already exists, and it cannot be upgraded to the requested schema ('+error+').',
+                    keyspace: req.keyspace,
+                    schema: newSchema
+                }
+            });
+        }
+        migrationPromise = migrationPromise.then(migrator.migrate.bind(migrator));
+    }
+
+    // Finally, if there are any migrations, apply them and persist.
+    if (migrationPromise !== emptyPromise) {
+        return migrationPromise
+            .then(function() {
+                var putReq = req.extend({
+                    columnfamily: 'meta',
+                    schema: self.infoSchemaInfo,
+                    query: {
+                        attributes: {
+                            key: 'schema',
+                            value: newSchema
+                        }
+                    }
+                });
+                return self._put(putReq)
+                    .then(function() {
+                        // Force a cache update on subsequent requests
+                        self._initSchemaCache();
+                        return { status: 201 };
+                    })
+                    .catch(function(error) {
+                        self.log('error/cassandra/table_update', error);
+                        throw error;
+                    });
+            });
+    }
+    else {
+        return {
+            status: 201
+        };
+    }
+};
+
 DB.prototype.createTable = function (domain, query) {
     var self = this;
     if (!query.table) {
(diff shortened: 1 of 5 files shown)


lib/db.js
```diff
@@ -776,7 +761,18 @@ DB.prototype.createTable = function (domain, query) {
         var retries = 100; // We try really hard.
         var delay = 100; // Start with a 1ms delay
         function doCreateTables() {
-            return self._createKeyspace(req, replicationOptions)
+            var repFactor = req.query.options.durability === 'low' ? 1 : 3;
```
gwicke: We'll need similar logic in the schema migration code, to update replication settings for existing tables. Maybe move the replication option setup to a separate helper?

lib/db.js
```diff
@@ -961,6 +957,33 @@ DB.prototype.getTableSchema = function(domain, table) {
...
+        // Local data-center (from the perspective of the queried node).
+        return self.client.execute_p("SELECT data_center FROM system.local", [], options);
+    })
+    .then(function(res) {
+        res.rows.forEach(function(row) {
+            datacenters.add(row.data_center);
+        });
+
+        return Array.from(datacenters);
+    });
```
gwicke: As a sanity check, we should probably make sure that `self.conf.localDc` is always in this set. I suspect that will basically always be the case, but based on past experience with how the local system.peers table is (or isn't) updated, it shouldn't harm to make sure.

lib/db.js
eevans: I'm proposing that the ALTER be a manual process (or separately scripted), and that this only addresses the creation of new keyspaces.

Adding a new data-center should be a rare and exceptional event, and we probably want better control of the timing of the ALTER (more control than a single RESTBase restart).

lib/db.js
eevans: Has past experience ever indicated that a node (and it's corresponding data center) would be missing (versus having an entry for a node that is no longer there)?  I would think that would be a very grave error.

Either way, I'll add a check (but just hope it's use is limited to RESTBase misconfigurations, and not wildly inaccurate representations of network topology :)).

lib/db.js
gwicke: An ALTER of any kind is always bound to a deploy, so is effectively scripted.

Do you see a strong reason for treating replication setting updates differently from other schema migrations?

lib/db.js
gwicke: Yeah, catching RB misconfigurations should be good to. We are [passing the localDc to the driver](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/lib/index.js#L21) so that it can make proper routing decisions, so it better is correct.

lib/db.js
eevans: > I wonder if we should use the RESTBase config to manage this, rather than pulling the list of DCs from Cassandra. That would keep control over schema changes firmly in our hands, with nothing happening without a deploy.

This is definitely better (better than to have it occur immediately upon discovering a new data-center), but I'm still skeptical that it's wisest to handle this here.  If nothing else, the contract this establishes is that altering the configuration alters the replication factor, when in fact, a rebuild is required to obtain the requested redundancy for everything but subsequent writes.

gwicke: We discussed this on IRC last Friday, and agreed to drive this primarily from the config, very similar to other schema changes. There will be a need to run repairs manually, so we should clearly document the need to do so in log entries and comments. We should also add information about this in https://wikitech.wikimedia.org/wiki/RESTBase, or consider building up docs on RESTBase administration on mw.org.

gwicke:
@eevans: ping ;)

eevans:
> @eevans: ping ;)

Yup, it's next on my list!

gwicke:
@eevans: We'll need this soon for testing.

lib/db.js
```diff
@@ -687,6 +687,107 @@ DB.prototype._migrateBackend = function(req, from, to) {
...
+                req,
+                currentSchemaInfo,
+                newSchemaInfo);
+        }
+        catch (error) {
+            throw new dbu.HTTPError({
+                status: 400,
+                body: {
+                    type: 'bad_request',
+                    title: 'The table already exists, and it cannot be upgraded to the requested schema ('+error+').',
```
d00rman: Nit: spaces around `('+error+')`

lib/db.js
```diff
@@ -687,6 +687,107 @@ DB.prototype._migrateBackend = function(req, from, to) {
...
+                    schema: self.infoSchemaInfo,
+                    query: {
+                        attributes: {
+                            key: 'schema',
+                            value: newSchema
+                        }
+                    }
+                });
+                return self._put(putReq)
+                    .then(function() {
```
d00rman: Nit: indentation

lib/db.js
```diff
@@ -687,6 +687,107 @@ DB.prototype._migrateBackend = function(req, from, to) {
...
+                        self._initSchemaCache();
+                        return { status: 201 };
+                    })
+                    .catch(function(error) {
+                        self.log('error/cassandra/table_update', error);
+                        throw error;
+                    });
+            });
+    }
+    else {
```
d00rman: Nit: `else` goes on the same line as `}` above it.

lib/db.js
```diff
@@ -972,13 +975,34 @@ DB.prototype._dropDomainIndex = function(req) {
...
+DB.prototype._createKeyspace = function (req) {
     var cql = 'create keyspace if not exists ' + cassID(req.keyspace)
-        + ' WITH REPLICATION = ' + options;
+        + ' WITH REPLICATION = ' + this._createReplicationOptionsCQL(req.query.options);
     return this.client.execute_p(cql, [],
             {consistency: req.consistency || this.defaultConsistency});
 };
 
+DB.prototype._createReplicationOptionsCQL = function(options) {
+    var durability = (options && options.durability === 'low') ? 1 : 3;
```
d00rman: Could be worth putting this into its own function so as to minimise copy/pasting (given its equality with line 999 below).

lib/db.js
```diff
@@ -1014,6 +1038,84 @@ DB.prototype.getTableSchema = function(domain, table) {
...
+    return this.client.execute_p(cql, [], {consistency: this.defaultConsistency});
+};
+
+/**
+ * Evaluates whether current keyspace replication matches the policy (a function of
+ * the configured datacenters, and the requested durability); Updates replication
+ * if necessary.
+ *
+ * NOTE: All this does is ALTER the underlying Cassandra keyspace, a repair (or
+ * cleanup) is is still necessary.
```
d00rman: Nit: `is is`

lib/index.js
```diff
@@ -8,13 +8,26 @@ var DB = require('./db');
...
 P.promisifyAll(cass, { suffix: '_p' });
 
+function validateAndNormalizeDcConf (conf) {
+    // Default to 'datacenter1'
+    if (!conf.localDc) { conf.localDc = 'datacenter1'; }
+    if (!conf.datacenters) { conf.datacenters = ['datacenter1']; }
+    if (!(conf.datacenters instanceof Array)) {
+        throw new Error('invalid datacenters configuration (not an array)');
+    }
+    if (conf.datacenters.indexOf(conf.localDc) < 0) {
```
d00rman: Nice check!

test/functional/replication.js
```diff
@@ -0,0 +1,74 @@
+"use strict";
+
+// mocha defines to avoid JSHint breakage
+/* global describe, it, before, beforeEach, after, afterEach */
```
d00rman: Heh, we are blindly copy/pasting these but they are not really needed since we are ignoring the `test` directory altogether in `.jshintrc`

test/functional/replication.js
```diff
@@ -0,0 +1,74 @@
...
+        { attribute: 'tid', type: 'range', order: 'desc' }
+    ],
+};
+
+describe('Table creation', function() {
+    var db;
+    before(function() {
+        return makeClient({
+            log: function(level, info) {
+                if (!/^info|warn|verbose|debug|trace/.test(level)) {
```
d00rman: Minor: isn't this equal to `if (/^error|fatal/)` ? Might be clearer this way.

test/functional/replication.js
```diff
@@ -0,0 +1,74 @@
...
+                    console.log(level, info);
+                }
+            },
+            conf: yaml.safeLoad(fs.readFileSync(__dirname + '/../utils/test_client.conf.yaml'))
+        })
+        .then(function(newDb) {
+            db = newDb;
+        })
+        .then(function() {
```
d00rman: No need for the `.then()` here, its contents can be safely put in the one above AFAICT.

lib/db.js
```diff
@@ -687,6 +687,107 @@ DB.prototype._migrateBackend = function(req, from, to) {
...
+                }
+            });
+        }
+        migrationPromise = migrationPromise.then(migrator.migrate.bind(migrator));
+    }
+
+    // Finally, if there are any migrations, apply them and persist.
+    if (migrationPromise !== emptyPromise) {
+        return migrationPromise
+            .then(function() {
```
d00rman: Indentation as well :)

d00rman:
Modulo minor nits, :+1: Really nice stuff @eevans !

d00rman:
Oh, one other minor thing - bump the package's version too as it's a rather important change for us.

gwicke:
Looks good overall to me as well.

test/functional/replication.js
eevans: my editor uses jshint to highlight issues, without these ignores I get false positives.

PR #141, opened 2015-09-03
Added support for static column additions

Pchelolo:
For page-related information we would need to be able to add static columns to the schema. This PR adds this capability to Cassandra and SQLite.

This PR depends on tests in https://github.com/wikimedia/restbase-mod-table-spec/pull/14


lib/schemaMigration.js
```diff
@@ -88,7 +93,7 @@ Attributes.prototype.migrate = function() {
     return P.each(self.addColumns, function(col) {
         self.log('warn/schemaMigration/attributes', {
             message: 'adding column' + col,
-            column: col,
+            column: col
```
gwicke: Heh, you are removing all my trailing commas ;)

PR #146, opened 2015-09-11
Added support for secondary index removal

Pchelolo:
There was an idea to drop a by_ns secondary index, this adds a test for this feature in backends.

I'm wondering, if we should actually drop the table, or just store an updated schema? I'd better alter this PR and do the latter, to have 100% protection agains data corruption. And the tables could then be deleted manually. What do you think?

Bug: https://phabricator.wikimedia.org/T111959

### DIFF
--- a/lib/schemaMigration.js
+++ b/lib/schemaMigration.js
@@ -208,10 +208,10 @@ function SecondaryIndexes(parentMigrator, current, proposed) {
 SecondaryIndexes.prototype.validate = function() {
     var self = this;
     if (self.addedIndexes.length > 0) {
-        throw new Error('Secondary index addition not supported');
+        throw new Error('Adding secondary indices is not supported');
     }
     if (self.changedIndexes.length > 0) {
-        throw new Error('Secondary index changing not supported');
+        throw new Error('Altering of secondary indices is not supported');
     }
 };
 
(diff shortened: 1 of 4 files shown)


lib/schemaMigration.js
```diff
@@ -183,10 +184,57 @@ Index.prototype.migrate = function() {
...
+    if (self.changedIndexes.length > 0) {
+        throw new Error('Secondary index changing not supported');
+    }
+};
+
+SecondaryIndexes.prototype.migrate = function() {
+    var self = this;
+    return self.deletedIndexes.forEach(function(indexName) {
+        self.log('warn/schemaMigration/secondaryIndexes', {
+            message: 'adding column' + indexName,
```
eevans: I'm guessing this message is copy-pasta; Maybe change this to 'deleting secondary index ...', instead?

lib/schemaMigration.js
```diff
@@ -183,10 +184,57 @@ Index.prototype.migrate = function() {
...
 }
 
-util.inherits(SecondaryIndexes, Unsupported);
+SecondaryIndexes.prototype.validate = function() {
+    var self = this;
+    if (self.addedIndexes.length > 0) {
+        throw new Error('Secondary index addition not supported');
+    }
+    if (self.changedIndexes.length > 0) {
+        throw new Error('Secondary index changing not supported');
```
eevans: <nit>How about, 'Altering of secondary indices is not supported' instead?</nit>

eevans:
LGTM overall, see inline comments for a few nits though.

PR #147, opened 2015-09-14
make TLS connections to Cassandra configurable

eevans:
https://phabricator.wikimedia.org/T111113

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -54,3 +54,111 @@ In production since March 2015.
 - [Leverage Cassandra 3 materialized
     views](https://phabricator.wikimedia.org/T111746) where it makes sense,
     once those have stabilized.
+
+## Configuration
+Configuration of this module takes place from within an `x-modules` stanza in the YAML-formatted
+[RESTBase configuration file](https://github.com/wikimedia/restbase/blob/master/config.example.wikimedia.yaml).
+While complete configuration of RESTBase is beyond the scope of this document, (see the
+[RESTBase docs](https://github.com/wikimedia/restbase) for that), this section covers the
+[restbase-mod-table-cassandra](https://github.com/wikimedia/restbase-mod-table-cassandra) specifics.
+
+```yaml
+    - name: restbase-mod-table-cassandra
+      version: 1.0.0
+      type: npm
+      options: # Passed to the module constructor
+        conf:
+          hosts: [localhost]
+          username: cassandra
+          password: cassandra
+          defaultConsistency: localOne
+          localDc: datacenter1
+          datacenters:
+            - datacenter1
+          storage_groups:
+            - name: default.group.local
+              domains: /./
+```
+
+### Hosts
+A list of Cassandra nodes to use as contact points.
+
+```yaml
+    hosts:
+      - cassandra-01.sample.org
+      - cassandra-02.sample.org
+      - cassandra-03.sample.org
+```
+
+### Credentials
+Password credentials to use in authenticating with Cassandra.
+
+*Note: Optional; Leave unconfigured if Cassandra authentication is not enabled.*
+
+```yaml
+    username: someuser
+    password: somepass
+```
+
+### Default Consistency
+The Cassandra consistency level to use when not otherwise specified.  Valid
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -20,6 +21,28 @@ function validateAndNormalizeDcConf (conf) {
...
+    if (sslConf.key)  { sslOpts.key  = fs.readFileSync(sslConf.key);  }
+
+    if (sslConf.ca) {
+        sslOpts.ca = [];
+        if (sslConf.ca instanceof Array) {
+            sslConf.ca.forEach(function(ca) {
+                sslOpts.ca.push(fs.readFileSync(ca));
+            });
+        }
+        else {
```
d00rman: Nit: put in the line above

lib/index.js
```diff
@@ -20,6 +21,28 @@ function validateAndNormalizeDcConf (conf) {
     }
 }
 
+// sync
+function sslOptions(sslConf) {
+    var sslOpts = {};
+
+    if (sslConf.cert) { sslOpts.cert = fs.readFileSync(sslConf.cert); }
+    if (sslConf.key)  { sslOpts.key  = fs.readFileSync(sslConf.key);  }
```
d00rman: Nit: maybe enforce both of these to be strings?

gwicke:
Could you add docs for this in README ?

lib/index.js
eevans: Hrmm, we could (and I certainly will if you feel strongly), but I wonder if the resulting exceptions will add much over simply allowing the exception from `fs#readFileSync` to trickle up.

eevans:
> Could you add docs for this in README ?

Yeah, let's do that.

Readme.md
```diff
@@ -54,3 +54,95 @@ In production since March 2015.
...
     once those have stabilized.
+
+## Configuration
+Configuration of this module takes place from within an `x-modules` stanza in the YAML-formatted
+[RESTBase configuration file](https://github.com/wikimedia/restbase/blob/master/config.example.wikimedia.yaml).
+While complete configuration of RESTBase is beyond the scope of this document, (see the
+[RESTBase docs](https://github.com/wikimedia/restbase) for that), this section covers the
+[restbase-mod-table-cassandra](https://github.com/wikimedia/restbase-mod-table-cassandra) specifics.
+
+    - name: restbase-mod-table-cassandra
```
Pchelolo: Super-minor: if you wrap it to "```yaml", github would highlight the syntax

    - name: restbase-mod-table-cassandra

vs
```yaml
   - name: restbase-mod-table-cassandra
```

eevans: > if you wrap it to "```yaml", github would highlight the syntax

Shiny; Thanks Petr!

lib/index.js
gwicke: Marko might means specifying an encoding parameter, so that fs.readFileSync returns a string instead of a Buffer.

lib/index.js
```diff
@@ -39,6 +61,8 @@ function makeClient (options) {
         reconnection: new reconnection.ExponentialReconnectionPolicy(100, 2000, true)
     };
 
+    if (conf.tls) { clientOpts.sslOptions = sslOptions(conf.tls); }
```
gwicke: Since exceptions here will be sync, it would be good to wrap this into

```javascript
try {
   clientOpts.sslOptions = sslOptions(conf.tls);
} catch (e) {
  return P.reject(e);
}
```

Readme.md
```diff
@@ -54,3 +54,111 @@ In production since March 2015.
 - [Leverage Cassandra 3 materialized
     views](https://phabricator.wikimedia.org/T111746) where it makes sense,
     once those have stabilized.
+
+## Configuration
```
gwicke: Very nice!

gwicke:
LGTM overall, only one small error handling nit.

lib/index.js
gwicke: The node docs mention that either a string *or* a buffer are fine, so imho this is okay.

PR #150, opened 2015-09-17
Removed actual deleting of secondary index tables

Pchelolo:
We cannot actually drop a secondary index tables during the deploy, as other nodes would fail to update the secondary index. So, just update the metadata for now.


d00rman:
LGTM. Bump the version number as well so we can get it in production

PR #151, opened 2015-09-17
Cache replication factor updates

gwicke:
Only update replication factor checks / updates once per keyspace. Without
this caching and given a long list of domains, startup takes on the order of
minutes *per worker*.


eevans:
LGTM

PR #153, opened 2015-09-18
DB schema caching & migration cleanup

gwicke:
- Remove the keyspaceSchemaCache, and make the schemaCache purely
  keyspace-based. Document caches better.
- Encapsulate the keyspaceSchemaCache handling in _keyspaceName().
- Don't wipe all caches when creating a new table, or migrating a schema.
- Retry transient errors in schema retrieval when a second request indicates
  that the keyspace and table now exists.
- Improve logging in schema migrations.
- Avoid allocations and delete in schema hashing code.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -725,11 +725,11 @@ DB.prototype._migrateIfNecessary = function(req, currentSchemaInfo, newSchema, n
     // First, check if the replication factors need to be updated in line with
     // the cassandra module configuration.
     if (!this.replicationUpdateCache[req.keyspace]) {
+        self.log('info/cassandra/replication_update_check',
+                'Checking replication factor: ' + req.keyspace);
         migrationPromise = self._updateReplicationIfNecessary(req.domain,
                 req.query.table, req.query.options)
         .then(function() {
-            self.log('info/cassandra/replication_update_check',
-                    'Checking replication factor: ' + req.keyspace);
             // Remember the successful replication check / update
             self.replicationUpdateCache[req.keyspace] = true;
         });
--- a/lib/db.js
+++ b/lib/db.js
@@ -21,7 +21,7 @@ function DB (client, options) {
     // cassandra client
     this.client = client;
 
-    this._initSchemaCache();
+    this._initCaches();
 
     /* Process the array of storage groups declared in the config */
     this.storageGroups = this._buildStorageGroups(options.conf.storage_groups);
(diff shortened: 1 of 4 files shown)


d00rman:
I'm witnessing the same failures as Travis does for 0.12 and iojs (w4m for 0.10). So it might be there's something missing? I haven't had time to debug it properly.

Pchelolo:
@gwicke @d00rman I have the tests failing too. A bit of debugging shows an error: ` Cannot assign to read only property '_backend_version' of testTable0` in `dbu.makeSchemaHash()` as if it was defined with `Object.defineProperty`

d00rman:
> I have the tests failing too. A bit of debugging shows an error: Cannot assign to read only property '_backend_version' of testTable0 in dbu.makeSchemaHash() as if it was defined with Object.defineProperty

Ah @Pchelolo you're faster :) I was about to post the same discovery :P

d00rman:
Restoring the original version of `dbu.makeSchemaHash()` makes the tests pass correctly.

gwicke:
This is odd. Are we using [Object.freeze()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) on schemas?

gwicke:
The failure was caused by schemaMigrations using makeSchemaHash for random attributes. Using the stable stringify directly there avoids this issue.

eevans:
There is a lot going on here (and in touchy areas of the code), but LGTM

lib/db.js
```diff
@@ -701,7 +719,21 @@ DB.prototype._migrateIfNecessary = function(req, currentSchemaInfo, newSchema, n
...
+    var schemaWriteNeeded = false;
+
+    // First, check if the replication factors need to be updated in line with
+    // the cassandra module configuration.
+    if (!this.replicationUpdateCache[req.keyspace]) {
+        migrationPromise = self._updateReplicationIfNecessary(req.domain,
+                req.query.table, req.query.options)
+        .then(function() {
+            self.log('info/cassandra/replication_update_check',
+                    'Checking replication factor: ' + req.keyspace);
```
eevans: This log message is going to be a little misleading, since it comes after the replication check (and possibly even after the update, which is also logged).

eevans:
Other than the nit mentioned in-line, still LGTM.

lib/db.js
gwicke: Good point. Let me move it up.

PR #154, opened 2015-09-22
Added a check for non-existing columns on insert

Pchelolo:
When we try to insert a field that's not in the schema, cassandra silently ignores that and SQLite is throwing a non-descriptive error. However, extra fields clearly indicate a bug in the client - either on table schema, or on a query, so we need to fail-fast and throw an error.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -468,6 +468,12 @@ DB.prototype._put = function(req) {
     }
     var schema = req.schema;
     var query = req.query;
+    var missingColumn = Object.keys(query.attributes).find(function(attrName) {
+        return !schema.attributes[attrName];
+    });
+    if (missingColumn) {
+        throw new Error('Unknown attribute ' + missingColumn);
+    }
 
     var tid = query.attributes[schema.tid];
     if (!tid) {
--- a/lib/db.js
+++ b/lib/db.js
@@ -501,7 +501,7 @@ DB.prototype._put = function(req) {
             });
             idxReq.query = Object.assign({}, idxReq.query);
             idxReq.query.attributes = newQueryAttributes;
-            
+
             batch.push(dbu.buildPutQuery(idxReq));
         }
     }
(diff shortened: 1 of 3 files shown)


lib/db.js
```diff
@@ -468,6 +468,12 @@ DB.prototype._put = function(req) {
     }
     var schema = req.schema;
     var query = req.query;
+    var missingColumn = Object.keys(query.attributes).find(function(attrName) {
```
gwicke: We already perform a lot of checking in buildPutQuery. Maybe we could piggy-back on that?

Pchelolo: @gwicke we use `dbu.buildPutQuery` for secondary index updates, and for them we send-in columns that are not in an index table. So in that place such check would fail.

I'm actually not sure what's better on the performance side - construction of a correct set of attributes for secondary index update, or sending everything in cassandra and relying on it to ignore additional attributes..

gwicke: We use buildPutQuery for both the indexes [and the main data update](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/lib/db.js#L498). We also [already check whether attributes are in the schema](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/lib/dbutils.js#L693-L705). I think the main bit that's missing is to throw an error, rather than silently dropping extra attributes.

Pchelolo:
@gwicke I've updated the PR, so now we check for extra attributes in `buildPutQuery`, however now we have to modify the secondary index update code a bit, so that it doesn't send in all attributes, but only those present in a secondary index table.

lib/dbutils.js
```diff
@@ -704,6 +704,8 @@ dbu.buildPutQuery = function(req, noConvert) {
                 params.push(val);
             }
             placeholders.push('?');
+        } else if (!/^_ttl.*/.test(key)) { // Allow TTL fields not in the schema
```
d00rman: This will fire when `val === undefined && schema.attributes[key]`, so the check should be:

```js
else if (!/^_ttl.*/.test(key) && !schema.attributes[key])
```

PR #155, opened 2015-09-30
Start testing with node 4.1

gwicke:


d00rman:
Hmm, all flavours fail the *fails to insert unknown field* test. Nooot good

d00rman:
Ah that's because of #154

PR #156, opened 2015-10-06
Implemented support for a revisionRetentionPolicy with 0 count

Pchelolo:
In cassandra we start looking at a `revisionRetentionPolicy` starting from a row just after the row we've just inserted. This needs to be updated to support policy with `count === 0`, that just set a graceTTL.

Also, there's been a bug in buldGetQuery - when `withTTLs` option was provided it failed with a 500 error.

SQLite module doesn't need updates, spec module is updated with a test.

### DIFF
--- a/lib/revisionPolicy.js
+++ b/lib/revisionPolicy.js
@@ -16,8 +16,6 @@ function RevisionPolicyManager(db, request, schema) {
     this.policy = schema.revisionRetentionPolicy;
     this.request = dbu.makeRawRequest(request, { ttl: this.policy.grace_ttl });
     this.noop = this.policy.type === 'all';
-    // The latest / just inserted row is not processed by the retentionPolicy,
-    // so start with count 1.
     this.count = 0;
     if (this.policy.type === 'interval') {
         var interval = this.policy.interval * 1000;
--- a/lib/revisionPolicy.js
+++ b/lib/revisionPolicy.js
@@ -18,7 +18,7 @@ function RevisionPolicyManager(db, request, schema) {
     this.noop = this.policy.type === 'all';
     // The latest / just inserted row is not processed by the retentionPolicy,
     // so start with count 1.
-    this.count = 1;
+    this.count = 0;
     if (this.policy.type === 'interval') {
         var interval = this.policy.interval * 1000;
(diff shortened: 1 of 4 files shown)


d00rman:
LGTM, but let's cc @wikimedia/services and wait for wikimedia/restbase-mod-table-spec#20 to get merged to restart the tests

d00rman:
@Pchelolo could you also update the dep version of `restbase-mod-table-spec` in `package.json` ?

Pchelolo:
@d00rman version updated, the tests pass.

lib/revisionPolicy.js
```diff
@@ -18,7 +18,7 @@ function RevisionPolicyManager(db, request, schema) {
     this.noop = this.policy.type === 'all';
     // The latest / just inserted row is not processed by the retentionPolicy,
     // so start with count 1.
-    this.count = 1;
+    this.count = 0;
```
d00rman: Nit: the above comment needs to be updated accordingly as well.

gwicke:
Technically this looks good to me. I like that the impact is fairly low, even if this is tricky code & I had to read it several times to convince myself that it's all correct.

The only small doubt I have is about the interface. Exposing this functionality as "latest 0" might not be immediately obvious to everybody. Giving it a more intuitive dedicated retention mode name ("time", "time-limited" or the like) might perhaps be clearer. Internally, we could still rewrite that to "latest 0", so the implementation wouldn't need to change much.

What do you think?

Pchelolo:
@gwicke I was thinking about a non-clear name too. Anyway, to implement a policy for stashing we need this code-change. 

The rewrite should happen in spec module, when a retention policy is validated&normalised. (https://github.com/wikimedia/restbase-mod-table-spec/blob/master/lib/validator.js#L107)

I like `time-limited` name more, with only option `grace_ttl`. If everyone would be OK with that naming, I'll prepare one more PR for spec module tomorrow (this one will be the same)

d00rman:
I've had the same doubts. I like the idea of simply exposing at as a different policy which equals to *latest 0*.

As for the name bike-shed, I'd prefer something like *temp* or the like, but we can address it elsewhere. Merging this one and publishing.

PR #159, opened 2015-10-21
Conf ver

eevans:
Adds a version attribute to the module configuration, and persists it to the meta.value schema structure as a hash-excluded private attribute (`_config_version`).  Replication changes are conditional on a monotonically increasing configuration version.

Bug: https://phabricator.wikimedia.org/T76494

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -114,10 +114,13 @@ dbu.makeValidKey = function makeValidKey (key, length) {
  */
 dbu.makeSchemaHash = function makeSchemaHash(schema) {
     var _backend_version = schema._backend_version;
-    // eliminate _backend_version from hash comparisons (see: DB#createTable)
+    var _config_version = schema._config_version;
+    // eliminate private versions from hash comparisons (see: DB#createTable)
     schema._backend_version = undefined;
+    schema._config_version = undefined;
     var hash = stable_stringify(schema);
     schema._backend_version = _backend_version;
+    schema._config_version = _config_version;
     return hash;
 };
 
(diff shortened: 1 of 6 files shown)


Pchelolo:
From a code perspective this looks good to me, however I do agree with a point you're making in the ticket - this is a second step on a road to special-cases-hell. 

As I understand, the `_backend_version` and `_config_version` doesn't belong to a table schema, it's an application-vide metadata, so in the perfect world we could have a special `app.meta` table to hold it in a single place. I don't propose to do such a drastic change now, but let's consider it if we add yet another exceptional case.

Overall, LGTM.

gwicke:
@Pchelolo, backend version migrations need to be tracked per-table, so a global table wouldn't work.

Pchelolo:
@gwicke actually, I don't understand why. If I get all the code correctly, after each deploy each table will have equal backend version (right now it's 1), and on `CURRENT_BACKEND_VERSION` increase, each table will go through the backend migration, and again, have equal value.

eevans:
@Pchelolo 
> @gwicke actually, I don't understand why. If I get all the code correctly, after each deploy each table will have equal backend version (right now it's 1), and on CURRENT_BACKEND_VERSION increase, each table will go through the backend migration, and again, have equal value.

Backend migrations are applied on a table-by-table basis though, and the `_backend_version` attribute is only updated after a successful migration.

FWIW, my suggestion was to move `_{backend,config}_version` out of the `meta.value` JSON blob, to separate, but still per-table attributes.

Pchelolo:
@eevans ah, ok, my suggestion was indeed not good.

Anyway, the current code looks good to me.

gwicke:
Another option we could consider is including all those versions in the quick-check hash. We are interested in speeding up the common case of startup with nothing having changed. In that case, only checking the hash will be slightly more efficient and possibly less messy than checking the versions separately.

The only downside I see in doing it that way is the possibility of a small slow-down when only the backend version or config version has changed, as all parts of the schema will still be checked for changes. I would be surprised if this difference mattered though, compared to the hundreds / thousands of checks without hash changes.

eevans:
@gwicke 
> Another option we could consider is including all those versions in the quick-check hash. We are interested in speeding up the common case of startup with nothing having changed. In that case, only checking the hash will be slightly more efficient and possibly less messy than special-casing the version checks.

That's basically the approach I was referring to [here](https://phabricator.wikimedia.org/T76494#1740006).  The problem is that for the table schema parts, you'll still need to compare the actual content of the persisted versus requested schema (sans the other two versions).  It could work, it just really doesn't feel like much of a win when weighing the additional complexity against optimizing away a couple of integer comparisons in the check.  YMMV though.

lib/dbutils.js
```diff
@@ -215,15 +218,20 @@ dbu.eachRow = function eachRow(client, query, params, options, handler) {
...
  */
-dbu.validateAndNormalizeSchema = function validateAndNormalizeSchema(schema) {
+dbu.validateAndNormalizeSchema = function validateAndNormalizeSchema(schema, configVer) {
     if (!schema._backend_version) {
         schema._backend_version = dbu.CURRENT_BACKEND_VERSION;
     }
+    if (configVer) {
+        schema._config_version = configVer;
+    }
     return validator.validateAndNormalizeSchema(schema);
```
gwicke: I'm actually wondering if we should add both `_backend_version` and `_schema_version` after `validator.validateAndNormalizeSchema(schema)`, to make sure that they don't go missing. We have a todo in the schema validator to validate other attributes (https://github.com/wikimedia/restbase-mod-table-spec/blob/master/lib/validator.js#L199), which could lead to that result.

gwicke:
See #166 for a rebased & refactored version of this.

gwicke:
Subsumed by #166.

PR #163, opened 2015-11-03
Support long type mapped to bigint

Pchelolo:
This PR adds mapping of `long` schema type to `bigint` cassandra type.
Clients supply longs as strings due to Javascript Number precision limitations.

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -411,7 +411,7 @@ dbu.conversions = {
     timeuuid: { read: toString() },
     uuid: { read: toString() },
     long: {
-        read: function(val) { return val.toString(); },
+        read: toString(),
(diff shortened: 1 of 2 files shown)


lib/dbutils.js
```diff
@@ -406,7 +409,11 @@ dbu.conversions = {
     blob: { write: encodeBlob },
     varint: { read: toNumber() },
     timeuuid: { read: toString() },
-    uuid: { read: toString() }
+    uuid: { read: toString() },
+    long: {
+        read: function(val) { return val.toString(); },
```
d00rman: Shorter: `read: toString()`

d00rman:
Published as [v0.8.8](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v0.8.8)

PR #165, opened 2015-11-06
Version migration: Bug fix: Compare the versions, not the objects

d00rman:
There was a bug in the Version migrator validation code, where instead of comparing only the versions of the current and proposed schemas, their entire objects were being compared, yielding an incorrect check.

Also, bump the version to v0.8.10

### DIFF
--- a/lib/schemaMigration.js
+++ b/lib/schemaMigration.js
@@ -286,7 +286,7 @@ function Version(parentMigrator, current, proposed) {
 
 // versions must be monotonically increasing.
 Version.prototype.validate = function() {
-    if (this.current >= this.proposed) {
+    if (this.current.version >= this.proposed.version) {
         throw new Error('new version must be higher than previous');
     }
 };
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.8.9",
+  "version": "0.8.10",
   "dependencies": {
     "bluebird": "~2.8.2",
     "cassandra-driver": "~2.2.1",


d00rman:
cc @wikimedia/services

lib/schemaMigration.js
```diff
@@ -286,7 +286,7 @@ function Version(parentMigrator, current, proposed) {
 
 // versions must be monotonically increasing.
 Version.prototype.validate = function() {
-    if (this.current >= this.proposed) {
+    if (this.current.version >= this.proposed.version) {
```
Pchelolo: OMG, Javascript, why can't you throw a runtime error on this crazy things?

Pchelolo: In case `proposed.version === undefined` this will return false.

d00rman: Good catch @Pchelolo !

d00rman: Ah, actually, no, this will never be true because it is added [during the normalisation](https://github.com/wikimedia/restbase-mod-table-spec/blob/9819ece92040626ddf044a99cbeef2814b78afac/lib/validator.js#L171-L173) of the proposed schema.

Pchelolo: Indeed. Hm, why did the [test](https://github.com/wikimedia/restbase-mod-table-spec/blob/master/test/functional/simple.js#L88) started to fail then? I could imagine why it should've been failing before the fix and start passing after, but vise-versa? 

Actually, the test doesn't look correct - there's a forgotten version update, and all it's doing it removing a row, which's supported by both backends.

Pchelolo:
Actually, this PR is not needed at all, as `this.current` and `this.proposed` are already a version - we supply individual migrators just with the pieces of info they need.

d00rman:
Yup, indeed @Pchelolo . The problems I'm witnessing locally with migration do not seem to be related to this specific check, this is just where the bug manifests itself. Closing the PR.

PR #166, opened 2015-11-07
Config version support & migration refactor

gwicke:

PR #167, opened 2015-11-10
Hints: Check cassandra version before applying DTCS hint

Pchelolo:
Travis has cassandra 2.0.9, which doesn't support DTCS, so before applying a `timeseries` update hint we need to check that cassandra supports it. Support was added in 2.0.11 and 2.1.1.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -29,6 +29,8 @@ function DB (client, options) {
     this.storageGroups = this._buildStorageGroups(options.conf.storage_groups);
     /* The cache holding the already-resolved domain-to-group mappings */
     this.storageGroupsCache = {};
+
+    this.cassandraVersion = this._getCassandraVersion();
 }
 
 DB.prototype._initCaches = function() {
@@ -39,7 +41,6 @@ DB.prototype._initCaches = function() {
     // keyspace -> boolean: replication factors checked / updated
     this.replicationUpdateCache = {};
 };
-
 
 /**
  * Set up internal request-related information and wrap it into an
@@ -97,7 +98,7 @@ DB.prototype._fetchSchema = function(req) {
             // Need to parse the JSON manually here as we are using the
             // internal _getRaw(), which doesn't apply transforms.
             var schema = JSON.parse(res.items[0].value);
-            schema = dbu.validateAndNormalizeSchema(schema, self);
+            schema = dbu.validateAndNormalizeSchema(schema);
             self.schemaCache[req.keyspace] = req.schema = dbu.makeSchemaInfo(schema);
         }
         return req;
@@ -874,7 +875,7 @@ DB.prototype.createTable = function (domain, query) {
         var currentSchemaInfo = req.schema;
 
         // Validate and normalize the schema
-        var newSchema = dbu.validateAndNormalizeSchema(req.query, self);
+        var newSchema = dbu.validateAndNormalizeSchema(req.query);
         var newSchemaInfo = dbu.makeSchemaInfo(newSchema);
 
         if (currentSchemaInfo) {
@@ -1012,7 +1013,7 @@ DB.prototype._createTable = function (req, schema, columnfamily) {
     cql += ['(' + hashBits.join(',') + ')'].concat(rangeBits).join(',') + '))';
 
     // Add options for compression & compaction
-    cql += " WITH " + dbu.getOptionCQL(schema.options || {});
+    cql += " WITH " + dbu.getOptionCQL(schema.options || {}, self);
 
     if (orderBits.length) {
         cql += ' and clustering order by ( ' + orderBits.join(',') + ' )';
--- a/lib/db.js
+++ b/lib/db.js
@@ -97,7 +97,7 @@ DB.prototype._fetchSchema = function(req) {
             // Need to parse the JSON manually here as we are using the
             // internal _getRaw(), which doesn't apply transforms.
             var schema = JSON.parse(res.items[0].value);
-            schema = dbu.validateAndNormalizeSchema(schema);
+            schema = dbu.validateAndNormalizeSchema(schema, self);
             self.schemaCache[req.keyspace] = req.schema = dbu.makeSchemaInfo(schema);
         }
         return req;
@@ -874,7 +874,7 @@ DB.prototype.createTable = function (domain, query) {
         var currentSchemaInfo = req.schema;
 
         // Validate and normalize the schema
-        var newSchema = dbu.validateAndNormalizeSchema(req.query);
+        var newSchema = dbu.validateAndNormalizeSchema(req.query, self);
         var newSchemaInfo = dbu.makeSchemaInfo(newSchema);
 
         if (currentSchemaInfo) {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -9,7 +9,8 @@
     "extend": "^2.0.0",
     "js-yaml": "^3.2.5",
     "json-stable-stringify": "git+https://github.com/wikimedia/json-stable-stringify#master",
-    "restbase-mod-table-spec": "^0.1.2"
+    "restbase-mod-table-spec": "^0.1.2",
```
d00rman: Could update the ver to 0.1.3

lib/dbutils.js
```diff
@@ -959,13 +960,20 @@ var updateOptionMap = {
         + " and gc_grace_seconds = 86400",
 };
 
-dbu.getTableCompactionCQL = function(options) {
+dbu.getTableCompactionCQL = function(options, cassandraVersion) {
```
d00rman: Since Cassandra's version is needed only here, maybe move `_getCassandraVersion()` to `dbutils.js` and call it directly from this method?

Pchelolo: We need a `client` object to get the version. Since `dbutils` doesn't have access to it, we'll need to pass in the `client` instead of a version. I thought passing in the version is a cleaner method contract

d00rman: Ah right, excellent point!

package.json
Pchelolo: @d00rman updated

lib/dbutils.js
```diff
@@ -959,13 +960,20 @@ var updateOptionMap = {
...
     var pattern = options.updates && options.updates.pattern;
     if (!pattern) {
         // Default to 'random-update'
         pattern = 'random-update';
     }
 
+    if (pattern === 'timeseries' && semver.valid(cassandraVersion)) {
+        // DTCS is supported starting from cassandra 2.0.11 or 2.1.1
+        if (!semver.satisfies(cassandraVersion, '>= 2.0.11 < 2.1.0 || >= 2.1.1')) {
+            pattern = 'random-update';
```
d00rman: Let's also warn about the fact that we are ignoring the user-supplied compaction strategy.

Pchelolo: @d00rman done. And also updated the migrator, which I forgot to updated in the previous version.

d00rman:
LGTM in general, but I want to raise two points:

- *Schema migrations*. Changing the compaction strategy **after** validation means that schemas will be incorrectly described in storage, possibly leading to conflicts when updating Cassandra and/or the scheme in question.
- *Unconditional strategy change*. This PR proposes to switch to a different, but supported, strategy in order to ensure the module's functioning. But, I'd much rather see this switch being a user-controlled conditional. To that end, a configuration option could be added to the module that enables/disables it.

Pchelolo:
@d00rman Agreed on both points. Now we fallback inside `validateAndNormalizeSchema`, it's a better place for that. Also, by default an error is thrown instead of a fallback.

gwicke:
I don't think we should throw errors while figuring out the best storage strategy for hints. It's not a config option demanding DTCS, after all; all it does it telling the storage backend to optimize storage for primarily `timeseries` data.

Pchelolo:
@gwicke Better now?

lib/db.js
```diff
@@ -950,6 +950,14 @@ DB.prototype.createTable = function (domain, query) {
     });
 };
 
+DB.prototype._getCassandraVersion = function() {
```
gwicke: Could you cache the return value, so that we only check this once?

lib/dbutils.js
```diff
@@ -221,10 +222,24 @@ dbu.CURRENT_BACKEND_VERSION = 1;
  * restbase-m-t-spec). Ensures the presence of the private,
  * implementation-specific _backend_version schema attribute.
  */
-dbu.validateAndNormalizeSchema = function validateAndNormalizeSchema(schema) {
+dbu.validateAndNormalizeSchema = function validateAndNormalizeSchema(schema, db) {
     if (!schema._backend_version) {
         schema._backend_version = dbu.CURRENT_BACKEND_VERSION;
     }
+    if (db) {
```
gwicke: This is the wrong place for this, imho.

lib/db.js
Pchelolo: Now it's set to a DB field

gwicke:
@Pchelolo, the caching looks good to me.

Could you take a look at https://github.com/wikimedia/restbase-mod-table-cassandra/pull/167#discussion-diff-44449666 as well?

Pchelolo:
@gwicke I've moved that away too already

gwicke:
@Pchelolo, this is still changing the schema, which is completely correct. Only the mapping from schema to Cassandra options should change, imho.

Pchelolo:
@gwicke @d00rman had a good point about that - in case we just change the mapping, and not the schema, when client updates the cassandra version to something that supports DTCS, migration will not fire. And if we change the schema, on next deploy to the updated infrastructure, schema migrations will fire up, and the fallback will get eliminated without any actions from the user.

gwicke:
Changing the schema on Cassandra upgrade can be good or bad. In large production environments like ours we don't want it to upgrade automatically, so updating the handling will do the right thing. In small environments, either strategy (DTCS or STCS) will work fine for timeseries data, and subsequent schema changes can still upgrade it where desired. 

Either way, changing the schema to lie about the update pattern seems like a bad idea. We don't do this for compression options or SQLite either, and I think shouldn't start to do so.

Pchelolo:
@gwicke actually in SQLite we completely ignore the compression and update-pattern options :)
Ok, I'll update one more time.

lib/dbutils.js
```diff
@@ -959,13 +963,22 @@ var updateOptionMap = {
...
     var pattern = options.updates && options.updates.pattern;
     if (!pattern) {
         // Default to 'random-update'
         pattern = 'random-update';
     }
 
+    if (pattern === 'timeseries' && db.cassandraVersion && semver.valid(db.cassandraVersion)) {
+        // DTCS is supported starting from cassandra 2.0.11 or 2.1.1
+        if (!semver.satisfies(db.cassandraVersion, '>= 2.0.11 < 2.1.0 || >= 2.1.1')) {
+            pattern = 'random-update';
```
gwicke: Make this write-once?

Pchelolo: @gwicke I would expect, that if some feature is not supported, the fallback would be the default value

gwicke: We seem to have the same argument over supported vs. not supported. My view is that the 'timeseries' update pattern is supported just fine. It just maps to different backend options depending on the backend used. In the case of old Cassandra and a timeseries update pattern, the best option seems to be STCS, which is why I proposed it.

gwicke:
LGTM. @eevans, could you have a look too?

lib/schemaMigration.js
```diff
@@ -41,6 +41,7 @@ function Options(parentMigrator, current, proposed) {
     this.current = current;
     this.proposed = proposed;
 
+    this.parent = parentMigrator;
```
eevans: Is this right?  It looks like you need the `DB` object for `dbu#getOptionCQL`, isn't that `parentMigrator.db`?

gwicke: Yeah, I think that's the case. This will also need to change with https://github.com/wikimedia/restbase-mod-table-cassandra/pull/166.

gwicke: Pushed a fix in #168.

PR #170, opened 2015-11-10
Test with node 4.2 instead of iojs & 4.1

gwicke:


eevans:
:+1:

gwicke:
Merged per @eevans :+1 & it being a trivial change that unbreaks tests (travis doesn't support 4.1 any more).

PR #172, opened 2015-12-03
Fixed TTL revPolicy and removed support for individual rows ttl

Pchelolo:
There's a whole bunch of bugs being fixed here:
- The TTL policy didn't work on cassandra, but that was unnoticed because of a bug in the test. The problem is that TTL is column-based in cassandra, so to actually set a TTL when writing a row, we need to have at least one non-null non-primary-key value in the row. But in revPolicy update code, we were fetching only key columns and those needed for secondary indexes, so for tables with no sec indexes TTL was not set up. Easy to fix by dropping the custom proj and fetching everything. We don't loose to much on the speed here, because it affects only a single row.
- After that being fixed, I've run into a bug in cassandra, that affects our tests. The bug is present in 2.1.* but fixed in 2.2.*. The problem was, that when you add a static column to an existing table, and try to fetch values of this column using `order by` query, the requests timeout. This is a problem for us when the table has a revPolicy and we add a static column there. Workaround for this bug is too complicated, so I've left everything as is, just modified the tests to not run into this problem.

The corresponding changes of the spec module are here: https://github.com/wikimedia/restbase-mod-table-spec/pull/29
The tests would fail until the spec changes are published.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.8.13",
+  "version": "0.8.14",
   "dependencies": {
     "bluebird": "~2.8.2",
     "cassandra-driver": "^2.2.2",
--- a/lib/db.js
+++ b/lib/db.js
@@ -54,7 +54,7 @@ DB.prototype._makeInternalRequest = function(domain, table, query, consistency) 
         consistency = cass.types.consistencies[query.consistency];
     }
     var keyspace = this._keyspaceName(domain, table);
-    var opts = {
+    var req = new InternalRequest({
         domain: domain,
         table: table,
         keyspace: keyspace,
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.8.12",
+  "version": "0.8.13",
```
d00rman: s/13/14/

gwicke:
Re-running tests after merging & publishing wikimedia/restbase-mod-table-spec#29.

d00rman:
Published as [v0.8.14](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v0.8.14)

d00rman:
After publishing this version, [this test](https://github.com/wikimedia/restbase/blob/8cb51a6e69fb3f0839959fbcda7c779cd5c9c216/test/features/errors/notfound.js#L83) started [failing for RESTBase](https://travis-ci.org/wikimedia/restbase/jobs/102247522) only when the Cassandra back-end is used. Version 0.8.13 allows it to [pass the tests](https://travis-ci.org/d00rman/restbase/jobs/102288362) again, so we should probably revisit this PR.

PR #174, opened 2015-12-09
Don't require early returns

gwicke:
Latest jscs seems to have started to disallow constructs of this form by
default:

if (something) {
    return a;
} else {
    return b;
}

I personally think this form is fine, especially if the branches are short.
The existing code also uses this quite often.


Pchelolo:
LGTM.

gwicke:
Merging per @Pchelolo and https://github.com/wikimedia/restbase/pull/439.

PR #175, opened 2015-12-09
Update parameters for DTCS tables & improve backend migration

gwicke:
This patch updates DTCS parameters for current stable Cassandra (2.2.4 and
2.1.12) by setting max_window_size_seconds explicitly, and dropping some less
important parameters. The exact settings aren't completely final yet, so we
might want to wait with applying this for a week or two until we have gotten a
bit more data on good defaults. These settings in particular are somewhat
questionable:
- unchecked_tombstone_compaction is disabled on some high-traffic prod tables,
  but we might want to revisit this in light of 2.1.12.
- tombstone_threshold: Set to 0.15 on high-traffic prod tables, but might also
  want to lower this in light of the increased compaction headroom in
  2.1.12. The current patch specified 0.1, which might be a good balance.
- max_window_size_seconds is set to 60 days in large production tables & this
  patch. We might want to increase this.

This patch also provides a backend migration pass to apply these parameter
changes. To make it easy to call other schema migrators, the main
BackendMigrator code now has all the logic needed.

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -944,7 +944,7 @@ var updateOptionMap = {
         // http://thread.gmane.org/gmane.comp.db.cassandra.devel/9753 and
         // http://mail-archives.apache.org/mod_mbox/cassandra-commits/201405.mbox/%3C8eaf7fbf908a4e40a77bc26b69ac7867@git.apache.org%3E
         + "'tombstone_threshold': '0.1', 'unchecked_tombstone_compaction': 'true', "
-        // Set the max window size (sstables to compact together) to three months.
+        // Set the max window size (sstables to compact together) to two months.
         + "'max_window_size_seconds': '5184000'}"
         // TODO: Calculate this based on the retention policy?
         + " and gc_grace_seconds = 86400",
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -200,7 +200,7 @@ dbu.eachRow = function eachRow(client, query, params, options, handler) {
  */
 
 dbu.DEFAULT_BACKEND_VERSION = 0;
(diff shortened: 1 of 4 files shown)


gwicke:
/cc @wikimedia/services 

Note that we might want to wait a little bit before merging this until we have a bit more data on the best DTCS defaults.

gwicke:
After running these settings for almost a week with reasonable results I'm confident that they are in the right ballpark for 2.1.12 and 2.2.4. Lets get this in soon.

lib/dbutils.js
```diff
@@ -937,18 +937,15 @@ var updateOptionMap = {
...
-        // - Lowered tombstone_threshold from 0.2 to 0.02
+        // - Lowered tombstone_threshold from 0.2 to 0.1
         // - Enabled unchecked_tombstone_compaction, which allows tombstone
         // collection even if key ranges overlap with other sstables. See
         // http://thread.gmane.org/gmane.comp.db.cassandra.devel/9753 and
         // http://mail-archives.apache.org/mod_mbox/cassandra-commits/201405.mbox/%3C8eaf7fbf908a4e40a77bc26b69ac7867@git.apache.org%3E
-        + "'tombstone_threshold': '0.02', 'unchecked_tombstone_compaction': 'true' }"
+        + "'tombstone_threshold': '0.1', 'unchecked_tombstone_compaction': 'true', "
+        // Set the max window size (sstables to compact together) to three months.
+        + "'max_window_size_seconds': '5184000'}"
```
Pchelolo: Minor: It's two months actually.

gwicke: Good catch. Fixed.

d00rman:
> Lets get this in soon.

:+1: Needs a rebase now, though.

PR #176, opened 2015-12-14
Yield on non-migration to avoid stack overflow

gwicke:


gwicke:
Self-merging, as quite trivial.

PR #177, opened 2016-01-14
Travis: Install custom cassandra version to match production

Pchelolo:
Tests the module against the same cassandra version we use in production. 

cc @wikimedia/services


d00rman:
Thnx @Pchelolo !

PR #179, opened 2016-01-27
TTLs: added support to get and set TTLs for individual rows

Pchelolo:
See https://github.com/wikimedia/restbase-mod-table-spec/pull/34

cc @wikimedia/services

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -54,7 +54,7 @@ DB.prototype._makeInternalRequest = function(domain, table, query, consistency) 
         consistency = cass.types.consistencies[query.consistency];
     }
     var keyspace = this._keyspaceName(domain, table);
-    var req = new InternalRequest({
+    var opts = {
         domain: domain,
         table: table,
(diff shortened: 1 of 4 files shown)


lib/db.js
```diff
@@ -238,7 +243,8 @@ DB.prototype.get = function(domain, query) {
     var self = this;
     return this._makeInternalRequest(domain, query.table, query)
     .then(function(req) {
-        return self._getRaw(req)
+        var options = query.withTTL ? { withTTLs: true } : undefined;
```
d00rman: Where is this option used?

Also, IMHO it'd be better to keep consistent with the API name (`withTTL`)

Pchelolo: @d00rman renamed the option. Now it should be obvious where is it used

d00rman:
Published as [v0.8.16](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v0.8.16)

PR #180, opened 2016-01-27
Don't mess with write timestamps and allow overwriting rows

Pchelolo:
In our system every db entry has a tid, which for us works as internal cassandra timestamp. We've used to assign the write timestamp the same as tid, so when we wanted to overwrite the value, there was a conflict and a new value was chosen by cassandra in unpredictable manner. 

Also, some miner improvements to revision policies are made: as we don't mess with timestamps any more, we rely on tid in the retention policy, which is actually way more correct.

Test here: https://github.com/wikimedia/restbase-mod-table-spec/pull/35

cc @wikimedia/services

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -415,8 +415,6 @@ DB.prototype._put = function(req) {
         query.attributes[schema.tid] = TimeUuid.fromString(query.attributes[schema.tid]);
     }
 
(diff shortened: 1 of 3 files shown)


d00rman:
Published as [v0.8.17](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v0.8.17)

PR #182, opened 2016-03-01
Rev Policy: Implemented latest_hash policy

Pchelolo:
See https://github.com/wikimedia/restbase-mod-table-spec/pull/36


Pchelolo:
Published as v0.8.18

PR #183, opened 2016-03-05
Performance: Switch to our own fork of cassandra-driver

gwicke:
We have a custom fork of the cassandra driver with several performance
improvements:

- Using `neo-async` instead of `async` reduces memory consumption
  significantly, which might eliminate worker restarts.
- Elimination of a couple of known-expensive operations
  (Object.defineProperty, bind) in a few strategic places lets V8 optimize the
  code better.

Overall, these changes yield about a 10% throughput improvement when tested
with
`ab -c20 -n 100000 http://localhost:7231/en.wikipedia.org/v1/page/html/FOO`.
The bigger impact is on memory however, with resident memory per worker
dropping from around 200m (and growing) to a stable ~118m.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,10 +1,10 @@
 {
   "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.8.18",
+  "version": "0.8.19",
   "dependencies": {
     "bluebird": "^3.1.1",
-    "cassandra-driver": "^2.2.2",
+    "cassandra-driver": "git+https://github.com/wikimedia/nodejs-driver#optimizations",
     "core-js": "^2.0.3",
     "extend": "^3.0.0",
     "js-yaml": "^3.5.2",


Pchelolo:
LGTM, however given the size of a change I think this could get a `0.9.0` version

d00rman:
I second @Pchelolo's comment. Also, I think it'd be preferable to publish our fork on npmjs, perhaps as `wmf-nodejs-driver` or similar, instead of using the branch directly.

gwicke:
I have a few more performance tweaks in https://github.com/wikimedia/restbase-mod-table-cassandra/pull/184. While profile information is clearly indicating that these tweaks are removing hot spots, I am less sure about the overall throughput. Local measurements are very noisy and susceptible to context switches and temporary cpu frequency boosting. The best might be to repeat measurements in an environment like staging.

That said, I did get peak throughputs of 1600 req/s for small HTML requests ([[en:FOO]]) when using ScyllaDB, and 1400 req/s with Cassandra.

gwicke:
Regarding publishing this as a new package: I see this as a temporary exploration / solution, which doesn't really warrant creating a new npm package.

I think longer term we should look into switching to the latest version (3.x) of the driver. This will require some adjustments around the conversions (as those are now exposed by the driver), but will also prepare us for Cassandra 3. The risk is that there could be regressions in Cassandra 2.x support, as the main focus behind 3.x has been Cassandra 3.x support.

gwicke:
Closing, as this is superseded by #185.

PR #184, opened 2016-03-07
For discussion: Performance improvements

gwicke:


lib/db.js
```diff
@@ -529,7 +529,6 @@ DB.prototype._deleteOlderContent = function(req) {
             } else {
                 dataQuery.attributes[att] = { lt: req.query.attributes[att] };
             }
-            break;
```
Pchelolo: This was here for a reason and removing it is wrong. In case we have multiple non-tid range keys we want to include only the first one is the query. Without a break this will produce an invalid query in that case.

lib/dbutils.js
```diff
@@ -262,6 +269,7 @@ dbu.makeIndexSchema = function makeIndexSchema(dataSchema, indexName) {
             s.iKeyMap[att] = indexElem;
         }
     });
+    s.iKeys.sort();
```
Pchelolo: What's the point of sorting these?

lib/dbutils.js
```diff
@@ -468,7 +488,7 @@ dbu.makeSchemaInfo = function makeSchemaInfo(schema, isMetaCF) {
     }
 
     // Create summary data on the primary data index
-    psi.iKeys = dbu.indexKeys(psi.index);
+    psi.iKeys = dbu.indexKeys(psi.index).sort();
```
Pchelolo: Before it was really handy that `hash` keys go first, then `range` keys. Some code relied on that I think.

gwicke: Yeah, the code in https://github.com/wikimedia/restbase-mod-table-cassandra/pull/184#discussion-diff-55275632L532, which is fixed ;)

jorgebay:
Hey!
I saw your branch for the driver optimizations, it looks good, I wanted to comment but I don't know if this is the right place, sorry to barge in:

Most of the performance improvements for the Node.js driver in the optimizations branch are already part of the driver since v3.0, [you can see the tickets on JIRA](https://datastax-oss.atlassian.net/browse/NODEJS-198?filter=-4&jql=project%20%3D%20NODEJS%20AND%20fixVersion%20in%20(3.0.0%2C%203.0.0-rc1)%20AND%20labels%20%3D%20performance%20ORDER%20BY%20updated) and the [changelog](https://github.com/datastax/nodejs-driver/blob/master/CHANGELOG.md), that includes: https://github.com/datastax/nodejs-driver/pull/118 and https://github.com/datastax/nodejs-driver/pull/116 and additional optimizations like https://github.com/datastax/nodejs-driver/pull/108.

There is another improvement coming as part of https://github.com/datastax/nodejs-driver/pull/130: avoid buffer region copies on streams module when possible.

Let me know if I can help in any way. We are committed to provide the best performance for the driver. We included performance testing as part of the release process (apart from the endurance tests) since v3.0. and is important priority for us.

gwicke:
@jorgebay, thanks for chiming in! I did indeed see your changes before playing with the performance tweaks in the 2.2 driver. We will look into moving to the 3.x driver.

gwicke:
@jorgebay, I did test the latest version (3.0.1), and did indeed notice a performance gain over 2.2, both the original & patched version. However, 3.0.1 does use significantly more memory in my tests, about 320mb RSS compared to ~140mb RSS using the 2.2 driver.

Switching to neo-async improves the situation a bit, see https://github.com/datastax/nodejs-driver/pull/134 for a PR and details.

PR #185, opened 2016-03-25
Driver upgrade

gwicke:
/cc @wikimedia/services

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.8.18",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.8.18",
+  "version": "0.8.19",
```
Pchelolo: @gwicke May be better 0.9.0 to avoid automatic updates?

gwicke: Yeah, makes sense.

gwicke:
Published as v0.9.0.

PR #190, opened 2016-07-11
Hack thin-out script to remove everything older than Dec 2015

eevans:
This should delete everything older than midnight 2015-12-31.  I have tested it with the query commented out, and the output looks right.

Bug: https://phabricator.wikimedia.org/T140008

### DIFF
--- a/maintenance/thin_out_key_rev_value_data.js
+++ b/maintenance/thin_out_key_rev_value_data.js
@@ -14,6 +14,9 @@ var preq = require('preq');
 
 if (!process.argv[3]) {
     console.error('Usage: node ' + process.argv[1] + ' <host> <keyspace>');
+    console.error('Usage: node ' + process.argv[1] + ' <host> <keyspace> [token]');
+    console.error('Usage: node ' + process.argv[1] + ' <host> <keyspace> [<domain> <key>]');
+    console.error('Usage: node ' + process.argv[1] + ' <host> <keyspace> [pageState]');
     process.exit(1);
 }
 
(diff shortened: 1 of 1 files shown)


maintenance/thin_out_key_rev_value_data.js
```diff
@@ -147,12 +150,20 @@ function processRow (row) {
...
         // Don't delete the most recent render for this revision
         return P.resolve();
-    } else if ((counts.rev > 0 && counts.render > 0)
+    } else if (false && ((counts.rev > 0 && counts.render > 0)
         || (counts.rev === 0 && counts.render > 0
             // Enforce a grace_ttl of 86400
             && (Date.now() - row.tid.getDate()) > 86400000)
-        || (counts.rev > 0 && row.tid.getDate() <  Date.parse('2015-07-10T13:00-0700'))) {
-        console.log('-- deleting:', row._token.toString(), row.tid.getDate().toISOString(), keys.rev);
+        || (counts.rev > 0 && row.tid.getDate() <  Date.parse('2015-12-31T23:59-0000')))) {
```
gwicke: The deletion logic seems to be the same as below -- why duplicate it rather than just tweaking the condition?

eevans: Unless I've misinterpreted things, the middle block will thin-out all revisions _and_ remove everything older than the hard-coded date.  The new block is only removes older entries.  As I remembered it, we were originally trying to catch records that predated our use of retention policies.

gwicke: > Unless I've misinterpreted things, the middle block will thin-out all revisions and remove everything older than the hard-coded date. 

It is only deleting *old revisions* (`counts.rev > 0`) older than the cut-off date.

> The new block only removes older entries. As I remembered it, we were originally trying to catch records that predated our use of retention policies.

No, the timestamp based part was always about removing old revisions completely, even if the retention policy would have kept them.

PR #193, opened 2016-07-26
Do not include the full request body in errors

gwicke:
Reduce the verbosity of storage backend errors by not including full request
bodies in the error object. Instead, the body is serialized to a JSON string,
and only the first 200 bytes of this string are returned.

Task: https://phabricator.wikimedia.org/T141384

### DIFF
--- a/index.js
+++ b/index.js
@@ -90,7 +90,11 @@ RBCassandra.prototype.get = function(rb, req) {
                 title: 'Error in Cassandra table storage backend',
                 stack: e.stack,
                 err: e,
-                req: req
+                req: {
+                    uri: req.uri,
+                    headers: req.headers,
+                    body: req.body && JSON.stringify(req.body).slice(0,200)
+                }
             }
         };
     });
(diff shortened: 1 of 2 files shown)


Pchelolo:
LGTM

index.js
```diff
@@ -90,7 +90,11 @@ RBCassandra.prototype.get = function(rb, req) {
                 title: 'Error in Cassandra table storage backend',
                 stack: e.stack,
                 err: e,
-                req: req
+                req: {
+                    uri: req.uri,
+                    headers: req.headers,
+                    body: JSON.stringify(req.body || {}).slice(0,200)
```
Pchelolo: `req.body && JSON.stringify(req.body).slice(0, 200)`

index.js
```diff
@@ -90,7 +90,11 @@ RBCassandra.prototype.get = function(rb, req) {
                 title: 'Error in Cassandra table storage backend',
                 stack: e.stack,
                 err: e,
-                req: req
+                req: {
+                    uri: req.uri,
```
Pchelolo: `req.params`?

index.js
Pchelolo: heh you forgot this one

Pchelolo:
Published as `restbase-mod-table-cassandra@0.10.1`

PR #195, opened 2016-08-18
Keyspace name script

eevans:
Simple script to fetch the calculated Cassandra keyspace name for a given RESTBase config.


Pchelolo:
LGTM

PR #196, opened 2016-08-26
Added a script to thin out non-normalized titles from big tables

Pchelolo:
In case the row.key is different from it's normalised version - delete the whole row since it's not reachable in RESTBase any more.

cc @wikimedia/services

### DIFF
--- a/maintenance/thin_out_non_normalized_parsoid.js
+++ b/maintenance/thin_out_non_normalized_parsoid.js
@@ -1,0 +1,231 @@
+"use strict";
+
+/*
+ * Known Issues:
+ *   - The token must be supplied in the form --token=, to prevent yargs from
+ *     parsing negative values as a flag.
+ *
+ */
+
+
+var P = require('bluebird');
+var cassandra = P.promisifyAll(require('cassandra-driver'));
+var ctypes = cassandra.types;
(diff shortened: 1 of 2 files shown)


maintenance/thin_out_non_normalized_parsoid.js
```diff
@@ -0,0 +1,231 @@
...
+            // Delete it
+            log('Deleting:', row._token.toString(), row.key);
+            var delHtml = 'DELETE FROM ' + htmlTable + ' WHERE "_domain" = ? AND key = ?';
+            var delData = 'DELETE FROM ' + dataTable + ' WHERE "_domain" = ? AND key = ?';
+            var delOffsets = 'DELETE FROM ' + offsetsTable + ' WHERE "_domain" = ? AND key = ?';
+            var params = [row._domain, row.key];
+            var delQueries = [
+                { query: delHtml, params: params },
+                { query: delData, params: params },
+                { query: delOffsets, params: params }
```
Pchelolo: I'm wondering though where we wanna delete it from all tables right away, summary, mobile stuff etc?

d00rman: I think all of them should go.

PR #198, opened 2016-11-22
Added a script to reload data from old restrictions table to the new one

Pchelolo:
Script that I intend to run to reload the data from old `restrictions` table to the new one once we create it. Don't need to merge, just showing the code that will run.

To be run after https://github.com/wikimedia/restbase/pull/718 is deployed.

cc @wikimedia/services


Pchelolo:
Very much outdated. Not needed anymore. Closing.

PR #200, opened 2017-01-24
Optimise the TTL revision retention policy

Pchelolo:
For the TTL policy we can avoid doing 'write -> read -> update' pattern because we always know that the TTL will be set, so we can skip the normal revision policy update and write with TTL right away.

Bug: https://phabricator.wikimedia.org/T150703

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -375,6 +375,12 @@ class DB {
             });
         }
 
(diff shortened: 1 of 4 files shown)


d00rman:
Published as [v0.10.3](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v0.10.3)

PR #201, opened 2017-01-27
Update to a newer cassandra driver and use built-in promise support

Pchelolo:
The new driver version is released and it has built-in promise support - let's use it!

I've measured the performance locally a bit, but no significant difference was found, need to run a dump in staging to get more data.

Bug: https://phabricator.wikimedia.org/T156441

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -173,7 +173,7 @@ class DB {
                 return req;
             } else {
                 // Check if the meta column family exists
-                return this.client.execute_p('SELECT columnfamily_name FROM '
+                return this.client.execute('SELECT columnfamily_name FROM '
                     + 'system.schema_columnfamilies WHERE keyspace_name=? '
                     + 'and columnfamily_name=?', [req.keyspace, 'meta'])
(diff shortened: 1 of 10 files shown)


eevans:
@d00rman said:
> Nevertheless, because we are switching to a different syntax (and use different underlying methods), we should carefully test it in staging.

Yup!

@Pchelolo does it make sense to likewise convert the uses under `maintenance/`?

Otherwise, LGTM :+1:

d00rman:
> @Pchelolo does it make sense to likewise convert the uses under maintenance/?

It would be great to convert those too as we can then use them to "test a bit" before going full-on into the replacement in RB.

d00rman:
:+1: Test next week?

Pchelolo:
@d00rman Yea, after deploying the more urgent matters

Pchelolo:
Done. Updated the maintenance scripts as well (not tested though)

Pchelolo:
Published as restbase-mod-table-cassandra@0.11.0

PR #202, opened 2017-03-10
Log schema hash mis-matches / migrations at "warn" level; v0.11.1

gwicke:
Schema migrations are significant events and can take a long time, so
should be brought to operators' attention by logging them at the "warn"
level.


d00rman:
Published as [v0.11.1](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v0.11.1)

PR #203, opened 2017-06-13
Don't provide a client key space option

Pchelolo:
Since the new release of the Cassandra-driver the client `keyspace` option matters. Since we reuse the client for many keyspaces, we always provide the keyspace in the request, so we shouldn't provide the client keyspace option. Previously this option was ignored.

I'm not entirely sure what exactly is happening within the driver, but since we share a single client for queries to many keyspaces, we definitely shouldn't set this option.

cc @wikimedia/services


d00rman:
Superseded by #204

PR #204, opened 2017-06-15
Fix: Quote the keyspace name; release v0.11.2

d00rman:
The new version of the driver (v3.2.2) actively uses the keyspace name to determine better routes when executing queries, which it previously was not doing. Alas, all of the keyspace names we were providing it were actually wrong, as they were not surrounded by double quotes. This caused the driver to look up metadata in all of the contact points, which resulted in a spiral time-out and retries.

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -924,10 +924,11 @@ class DB {
      */
     _getReplication(domain, table) {
         const keyspace = this.keyspaceName(domain, table);
-        const ks = this.client.metadata.keyspaces[keyspace];
+        const ks = this.client.metadata.keyspaces[keyspace] ||
+            this.client.metadata.keyspaces[keyspace.slice(1, -1)];
         const datacenters = {};
         if (!ks) {
-            return datacenters;
+            return P.resolve(datacenters);
         }
         Object.keys(ks.strategyOptions).forEach((dc) => {
             datacenters[dc] = parseInt(ks.strategyOptions[dc]);
--- a/lib/db.js
+++ b/lib/db.js
@@ -210,7 +210,7 @@ class DB {
         const reversedName = name.toLowerCase().split('.').reverse().join('.');
         const prefix = dbu.makeValidKey(reversedName, Math.max(26, 48 - table.length - 3));
         // 6 chars _hash_ to prevent conflicts between domains & table names
-        const res = `${prefix}_T_${dbu.makeValidKey(table, 48 - prefix.length - 3)}`;
+        const res = `"${prefix}_T_${dbu.makeValidKey(table, 48 - prefix.length - 3)}"`;
         this.keyspaceNameCache[cacheKey] = res;
         return res;
     }
@@ -288,7 +288,7 @@ class DB {
         // }
 
         // Paging request:
-        const cassOpts = { consistency: req.consistency, prepare: true };
+        const cassOpts = { consistency: req.consistency, prepare: true, keyspace: req.keyspace };
 
         if (req.query.limit) {
             cassOpts.fetchSize = req.query.limit;
--- a/lib/index.js
+++ b/lib/index.js
@@ -3,8 +3,6 @@
 const P = require('bluebird');
 const cass = require('cassandra-driver');
 const fs = require('fs');
-const loadBalancing = cass.policies.loadBalancing;
-const reconnection = cass.policies.reconnection;
 const DB = require('./db');
 
 function validateAndNormalizeDcConf(conf) {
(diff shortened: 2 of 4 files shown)


d00rman:
cc @wikimedia/services

lib/db.js
```diff
@@ -210,7 +210,7 @@ class DB {
         const reversedName = name.toLowerCase().split('.').reverse().join('.');
         const prefix = dbu.makeValidKey(reversedName, Math.max(26, 48 - table.length - 3));
         // 6 chars _hash_ to prevent conflicts between domains & table names
-        const res = `${prefix}_T_${dbu.makeValidKey(table, 48 - prefix.length - 3)}`;
+        const res = `"${prefix}_T_${dbu.makeValidKey(table, 48 - prefix.length - 3)}"`;
```
gwicke: Instead of adding the quote work-around here, do you think we could do so as late as possible, when setting up cassOpts? This might also avoid the need to change cassID.

d00rman: We use the keyspace in a lot of places, I'm not sure hunting down everywhere where the quotes are needed would be an easy feat and sounds error-prone.

lib/index.js
```diff
@@ -45,20 +43,8 @@ function makeClient(options) {
...
-    clientOpts.policies = {
-        loadBalancing: new loadBalancing.TokenAwarePolicy(
-            new loadBalancing.DCAwareRoundRobinPolicy(conf.localDc)
-        ),
-        // Also see
-        // http://www.datastax.com/documentation/developer/nodejs-driver/2.0/common/drivers/reference/clientOptions.html
-        // Retry immediately, then delay by 100ms, back off up to 120s
-        reconnection: new reconnection.ExponentialReconnectionPolicy(100, 120000, true)
-    };
-
```
Pchelolo: Why are we removing this? Is that the default? Don't we want to keep it even if the default changes?

lib/db.js
```diff
@@ -924,10 +924,11 @@ class DB {
      */
     _getReplication(domain, table) {
         const keyspace = this.keyspaceName(domain, table);
-        const ks = this.client.metadata.keyspaces[keyspace];
+        const ks = this.client.metadata.keyspaces[keyspace] ||
+            this.client.metadata.keyspaces[keyspace.slice(1, -1)];
         const datacenters = {};
         if (!ks) {
-            return datacenters;
+            return P.resolve(datacenters);
```
Pchelolo: What's the point?

lib/db.js
```diff
@@ -924,10 +924,11 @@ class DB {
      */
     _getReplication(domain, table) {
         const keyspace = this.keyspaceName(domain, table);
-        const ks = this.client.metadata.keyspaces[keyspace];
+        const ks = this.client.metadata.keyspaces[keyspace] ||
+            this.client.metadata.keyspaces[keyspace.slice(1, -1)];
```
Pchelolo: If we are wrapping the key space name in double quotes always, what's the point of this fallback? If the fallback actually needed, that means some key space was not quoted and we will have troubles with the driver?

lib/db.js
```diff
@@ -288,7 +288,7 @@ class DB {
         // }
 
         // Paging request:
-        const cassOpts = { consistency: req.consistency, prepare: true };
+        const cassOpts = { consistency: req.consistency, prepare: true, keyspace: req.keyspace };
```
Pchelolo: If this is needed there's a ton of other places where we invoke "execute" and provide these options. We should add it there as well.

d00rman: It doesn't seem to work without this, but with it does ... Hm hm

lib/db.js
d00rman: Otherwise we get the `this._getReplication().then is not a function` error when this code path is taken.

lib/index.js
d00rman: Yes, these are now defaults in the driver, so no need to be explicit for them when we use all of the other defaults as well.

lib/db.js
d00rman: This is something that I noticed in the tests only, the schema migration tests fail without this because the test keyspace names don't need to be quoted.

d00rman: Honestly, I'm not quite sure if this is needed or not

d00rman:
Let's hold off this PR, I need to test stuff a bit more, now I'm not sure this is the solution :/

gwicke:
At a higher level, having to pass in a pre-quoted keyspace to the driver feels like a driver bug. In CQL, the quotes make sense, but when passing a raw keyspace name in an option imho it does not. Do you think it is worth checking whether we can teach the driver to work with plain (unquoted) keyspace names instead? It sounds like some quote stripping (and possibly adding) might be missing in the driver.

d00rman:
Nevermind, folks, this is not the solution to our problem. The driver does not actually have the keyspace names quoted.

PR #206, opened 2017-07-10
Storage redesign: Remove 2ary indexing and retention policies

eevans:
This (work-in-progress) branch uses an alternative package name (restbase-mod-table-cassandra-ng) so that it can live alongside the current production code as we make an incremental, phased roll out.

* Strips out support for soft deletes (ala the `_del` attribute)
* Strips out support for secondary indexes
* Strips out retention policies
* Fixes the tests accordingly


Pchelolo:
Looks good, but I agree it's better to delete everything we need to delete in a single PR

Pchelolo:
+1

PR #208, opened 2017-07-11
Range-supporting DB#delete method (hard deletes)

eevans:
Expose deletes.

Bug: https://phabricator.wikimedia.org/T169938


lib/db.js
```diff
@@ -331,7 +331,11 @@ class DB {
     }
 
     _delete(req) {
-        return this._put(req);
+        const queryOptions = { consistency: req.consistency, prepare: true };
+        const queryInfo = dbu.buildDeleteQuery(req);
+        return this.client.execute(queryInfo.cql, queryInfo.params, queryOptions).thenReturn({
+            status: 201
```
Pchelolo: According to the [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.7) for `DELETE` requests:

> A successful response SHOULD be 200 (OK) if the response includes an entity describing the status, 202 (Accepted) if the action has not yet been enacted, or 204 (No Content) if the action has been enacted but the response does not include an entity.

So in our case it should be 204

lib/dbutils.js
```diff
@@ -799,6 +799,26 @@ dbu.buildGetQuery = (req, options) => {
...
+ * CQL building for DELETE queries
+ * @param {InternalRequest} req
+ * @param  {object} options map
+ * @return {object} queryInfo object with cql and params attributes
+ */
+dbu.buildDeleteQuery = (req) => {
+    const schema = req.schema;
+    const query = req.query;
+    const attributes = query.attributes || {};
+    if (req.columnFamily !== 'meta') {
```
Pchelolo: Em, what? We shouldn't allow issuing deletes to `meta` table, it's our internal implementation detail.

test/unit/delete.js
```diff
@@ -51,12 +53,37 @@ describe('Delete', function() {
...
+                    table: 'even-simpler-table',
+                    options: {
+                        durability: 'low',
+                        compression: [
+                            {
+                                algorithm: 'deflate',
+                                block_size: 256
+                            }
+                        ]
+                    },
```
Pchelolo: Any reason to explicitly overwrite defaults?

test/unit/delete.js
```diff
@@ -51,12 +53,37 @@ describe('Delete', function() {
...
+                        {attribute: 'rev', type: 'range', order: 'desc'}
+                    ]
+                });
             });
         });
     });
 
     after(function() {
-        return db.dropTable('restbase.cassandra.test.local', 'simple-table');
+        return P.map(['simple-table', 'even-simpler-table'], (i) => {
```
Pchelolo: In general for one-liners we don't use brackets for the param or curly brackets

test/unit/delete.js
```diff
@@ -70,4 +97,51 @@ describe('Delete', function() {
             }
         });
     });
+
+    it('puts the lotion on its skin', () => {
```
Pchelolo: heh

test/unit/delete.js
```diff
@@ -70,4 +97,51 @@ describe('Delete', function() {
...
+        return P.map(Array.from(new Array(20), (x, i) => i), (rev) => {
+            return db.put('restbase.cassandra.test.local', {
```
Pchelolo: Idem regarding brackets

lib/db.js
eevans: Yup, this was copy-pasta; I meant to circle back around and find the right status.

lib/dbutils.js
eevans: Hrmm, we do use this interface to (for example), write to `meta`.  See [lib/db.js#L657](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/lib/db.js#L657), and the similar handling in [lib/dbutils.js#L656](https://github.com/wikimedia/restbase-mod-table-cassandra/blob/master/lib/dbutils.js#L656).  I can't think of any reason that we'd *need* to do deletes on `meta`, but I wasn't sure there was reason to exclude it either.

Do you think we should raise an exception instead?

Pchelolo: I think deleting from meta is dangerous enough to raise an exception. If we delete from meta we effectively make the key space completely unusable.

test/unit/delete.js
eevans: Nope, this was just more copy-pasta.  This whole test file will be deleted though once tests are in place in wikimedia/restbase-mod-table-spec (i was just using this to sort out the initial implementation).

test/unit/delete.js
eevans: Gotcha

test/unit/delete.js
eevans: :)

index.js
```diff
@@ -126,6 +127,31 @@ class RBCassandra {
         }));
     }
 
+    delete(rb, req) {
+        const domain = req.params.domain;
+        // XXX: Use the path to determine the primary key?
+        return this.store.delete(domain, req.body)
+        .thenReturn({
+            // created
```
d00rman: Nit: `// deleted` ?

index.js
```diff
@@ -126,6 +127,31 @@ class RBCassandra {
...
+            status: 204
+        })
+        .catch((e) => ({
+            status: 500,
+
+            body: {
+                type: 'delete_error',
+                title: 'Internal error in Cassandra table storage backend',
+                stack: e.stack,
```
d00rman: Why is the stack here explicitly when the error below ought to contain it as well?

lib/dbutils.js
```diff
@@ -799,6 +799,27 @@ dbu.buildGetQuery = (req, options) => {
     return { cql, params };
 };
 
+/**
+ * CQL building for DELETE queries
+ * @param {InternalRequest} req
+ * @param  {object} options map
+ * @return {object} queryInfo object with cql and params attributes
```
d00rman: The latter two do not seem to be function arguments.

index.js
eevans: That's a Good Question.  The short and unsatisfying answer is that it is that way to be consistent with the other request types; I don't know why they are that way.  I wonder if it doesn't end up output as `[object]` or something?

lib/dbutils.js
eevans: Hrmm, not sure what you mean.  The options map object argument is copy-pasta (I've removed that), is that what you mean?

index.js
d00rman: Ok, I see. Let's leave this be for now then and address all of the cases at once.

PR #209, opened 2017-07-17
Expose default_time_to_live

Pchelolo:
As we're removing the `revisionRetentionPolicy` from the storage module, we need another way to specify table-level TTLs. So expose the `default_time_to_live` configuration property.

This will also need some changes to the restbase key_rev_value.

Bug: https://phabricator.wikimedia.org/T170901
cc @wikimedia/services


d00rman:
:+1:

eevans:
👍

PR #210, opened 2017-07-18
Switch to system_schema.* keyspace for checking if the meta table exist

Pchelolo:
The `system.schema_columnfamilies` table doesn't exist any more, see https://issues.apache.org/jira/browse/CASSANDRA-10996

cc @wikimedia/services


eevans:
LGTM 👍

PR #211, opened 2017-07-18
Remove implicit _TID adding

Pchelolo:
This one is not quite ready, because we need to adapt the tests and we need to check that RESTBase will not be broken by this, so created a PR to engage the discussion early on. 

Currently we're adding the implicit `tid`(or `_tid`) to all the tables, so we never overwrite the values, we're writing the new value with a new TID and then apply some retention policy. However, since we've got rid of the revision policies, we don't really need to have TID in every schema - it's a too magical and not manageable without retention policies. 

With this change we will be able to convert the `key_value` bucket to use overwriting instead of a rev policy, also we'd be able to create the `timeline` indexes.

Thoughts?

Bug: https://phabricator.wikimedia.org/T170902
cc @wikimedia/services

### DIFF
--- a/index.js
+++ b/index.js
@@ -181,12 +181,10 @@ class RBCassandra {
         return this.store.getTableSchema(domain, req.params.table)
         .then((res) => ({
             status: 200,
-            headers: { etag: res.tid.toString() },
             body: res.schema
         }))
         .catch((e) => ({
             status: 500,
-
             body: {
                 type: 'schema_query_error',
                 title: 'Internal error querying table schema in Cassandra storage backend',
--- a/lib/db.js
+++ b/lib/db.js
@@ -2,7 +2,6 @@
 
 const P = require('bluebird');
 const cass = require('cassandra-driver');
-const TimeUuid = cass.types.TimeUuid;
 const extend = require('extend');
 const dbu = require('./dbutils');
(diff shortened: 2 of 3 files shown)


d00rman:
Should we make it an option and add them if the client wishes to? Keeping the TID would allow us to model simple retention policies right in the storage module, and I think that makes sense.

Pchelolo:
@d00rman Do we want to have any retention policies in the storage module at all?

d00rman:
> Do we want to have any retention policies in the storage module at all?

Not necessarily as such, but having this in the storage module sets the right expectations, in that the storage module should be able to handle expiring records. This simplifies reasoning on the higher (bucket) level, where adequate retention policies can be implemented (whether they will need the TTLs or not it's a different matter and not related to the storage module).

Pchelolo:
So is the consensus to move forward on this one? @gwicke @d00rman @eevans

eevans:
Yes, sorry; LGTM 👍

Pchelolo:
Ok, will make a spec change to make the tests actually pass.

Pchelolo:
Here're the changes to the tests: https://github.com/wikimedia/restbase-mod-table-spec/pull/42

PR #212, opened 2018-01-08
Rename to remove -ng suffix; Bump version to 0.12.2

eevans:
Bug: https://phabricator.wikimedia.org/T184199

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "restbase-mod-table-cassandra-ng",
+  "name": "restbase-mod-table-cassandra",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "restbase-mod-table-cassandra-ng",
+  "name": "restbase-mod-table-cassandra",
   "description": "RESTBase table storage on Cassandra",
-  "version": "0.12.1",
+  "version": "0.12.2",
```
d00rman: Idem as for the spec package wrt versioning.

PR #215, opened 2018-06-12
Support dry-run for schema migrations

Pchelolo:
This will dry-dun the actual Cassandra schema alternation.

cc @wikimedia/services

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -369,6 +369,7 @@ class DB {
                 db: this,
                 client: this.client,
                 log: this.log,
+                skip_schema_update: this.conf.skip_schema_update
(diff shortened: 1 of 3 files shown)


d00rman:
Do we need the `console.log()` in the PR, though? We can just run it locally and collect it manually.

Pchelolo:
We don't have console.log :) we have the normal trace logging. for local stuff we can use, for example, sampled_levels and the same trace logging.

d00rman:
Published as [v1.0.1](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v1.0.1)

PR #216, opened 2018-06-14
Remove automagic compression/compaction settings and migrations.

Pchelolo:
Tested locally that this indeed clears up meta tables from compression/compaction settings without issuing any table alternations. Would be good to deploy this together with a RESTBase PR that would clear up update patterns from the schemas.

cc @wikimedia/services


eevans:
This looks pretty good to me; I like it :+1:

d00rman:
Hm all tests seem to be failing now?

d00rman:
Published as [v1.1.0](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v1.1.0)

PR #218, opened 2018-07-05
Dependencies update

Pchelolo:
1. Update all dependencies, mainly cassandra-driver to 3.5.0
2. Adapt to newer eslint settings
3. Remove jscs and jshint, they both are replaced with eslint

cc @wikimedia/services


d00rman:
Hmm this probably shouldn't have been merged before we tested it in production.

PR #222, opened 2019-03-13
Replace `json-stable-stringify` with `fast-json-stable-stringify`

clarakosi:
fast-json-stable-stringify is being used because it does not have any dependencies and has been shown to be faster than json-stable-stringify

Bug: [T210426](https://phabricator.wikimedia.org/T210426)

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -6,7 +6,7 @@ const crypto = require('crypto');
 const extend = require('extend');
 const cass = require('cassandra-driver');
 const P = require('bluebird');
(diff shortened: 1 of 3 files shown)


d00rman:
LGTM, but let's also bump the version so that it gets picked up by restbase on the next deploy.

clarakosi:
> LGTM, but let's also bump the version so that it gets picked up by restbase on the next deploy.

Done.

d00rman:
Published as [v1.1.3](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v1.1.3)

PR #223, opened 2019-05-28
Stringify a Buffer if we want to write it into string type.

Pchelolo:
For RESTBase split some of the legacy tables that we want to reuse were created with string value type. For unusual content-types, like mml or svg we need to stringify the buffer that the backend is receiving.

### DIFF
--- a/lib/dbutils.js
+++ b/lib/dbutils.js
@@ -313,6 +313,17 @@ dbu.conversions = {
     long: {
         read: toString(),
         write(val) { return Long.fromString(val); }
+    },
+    // Don't stringify anything. It's ok to stringify a Buffer,
+    // but in general we need to catch errors if object or integer
+    // were provided instead of the string.
+    string: {
+        write: (val) => {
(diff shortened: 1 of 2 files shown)


lib/dbutils.js
```diff
@@ -313,6 +313,17 @@ dbu.conversions = {
...
         read: toString(),
         write(val) { return Long.fromString(val); }
+    },
+    // Don't stringify anything. It's ok to stringify a Buffer,
+    // but in general we need to catch errors if object or integer
+    // were provided instead of the string.
+    string: {
+        write: (val) => {
+            if (Buffer.isBuffer(val)) {
+                return val.toString;
```
d00rman: `s/toString/toString()/`

Pchelolo: Done. Sorry.

d00rman:
Published as [v1.1.4](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v1.1.4)

PR #226, opened 2019-06-06
Replaced Oracle Java 8 with OpenJDK 8

hknustwmf:
Travis builds are failing. Also added Node 10 to Travis config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
   - "6"
-  - "8"
+  - "10"
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,6 +2,7 @@ language: node_js
 node_js:
   - "6"
   - "8"
```
Pchelolo: We can remove 8, we are never going to use it.

.travis.yml
```diff
@@ -12,7 +13,7 @@ notifications:
 addons:
   apt:
     packages:
-    - oracle-java8-set-default
+    - openjdk-8-jre-headless
```
Pchelolo: right....

.travis.yml
hknustwmf: Just removed it

PR #227, opened 2019-06-07
Create keyspaces for all storage groups if * domain is provided

Pchelolo:
Now in case the request to create bucket came with a `*` domain, it will create the bucket for all the defined storage groups.

cc @wikimedia/services

### DIFF
--- a/lib/db.js
+++ b/lib/db.js
@@ -87,18 +87,22 @@ class DB {
     /**
      * Set up internal request-related information and wrap it into an
      * InternalRequest instance.
-     * @param  {string}  domain      in dot notation
-     * @param  {string}  table       logical table name
-     * @param  {Object}  query       query object
-     * @param  {Object}  consistency consistency level
+     * @param  {string}   domain            in dot notation
+     * @param  {string}   table             logical table name
+     * @param  {Object}   query             query object
+     * @param  {Function} [resolveKeyspace] keyspace name resolver
+     * @param  {Object}   [consistency]     consistency level
      * @return {Object}
      */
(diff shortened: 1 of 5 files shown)


lib/db.js
```diff
@@ -177,14 +181,24 @@ class DB {
      * @return {string}         Valid Cassandra keyspace key
      */
     keyspaceName(domain, table) {
-        const cacheKey = JSON.stringify([domain, table]);
+        const cacheKey = JSON.stringify(['domain', domain, table]);
```
d00rman: I think this should be optimised. We don't need a per-domain cache, only on a per-storage-group basis.

d00rman:
Oh, and the version should be bumped

lib/db.js
Pchelolo: mmm... maybe this is an overoptimization, but we now have both by-domain and by-group caching. Resolving a storage group could mean executing quite a lot of regexes..

d00rman: Oh, right. What I meant was that the domains cache should be the mapping to the storage group and from then onward all the cache lookups should be by storage group. Instead, I believe the caches now map domains directly to names and statements, which means more executions than really needed.

Pchelolo: mmm.. we didn't really change how the caches work, we only introduced one more cache `storageGroup => ksName` that we didn't have before. If I understand correctly what you're proposing is to change all other caches to be `storageGroup => whatever` - it's a good optimization, but not really a part of this PR

d00rman:
Published as [v1.2.0](https://github.com/wikimedia/restbase-mod-table-cassandra/tree/v1.2.0)

PR #228, opened 2019-08-20
Maintenance: Conform to the new RB config layout

d00rman:


Pchelolo:
We need smth like https://github.com/wikimedia/restbase/pull/1169 to fix travis first

d00rman:
> We need smth like [wikimedia/restbase#1169](https://github.com/wikimedia/restbase/pull/1169) to fix travis first

The code changed here is not even tested, so IMHO this can be reviewed without that (as in, that is separate concern and hence PR)

Pchelolo:
Maks sence

PR #229, opened 2019-08-21
Travis: Fix the JAVA_HOME env variable

d00rman:
Also, set the Cassandra version as an env variable so that we can easily change it later on.


d00rman:
Trivial, self-merging.

PR #230, opened 2019-09-27
Cache table creation operations; release v1.2.2

d00rman:
Since we execute identical PUT /bucket requests for all domains within
the same storage group, use a cache to speed up the start-up sequence.
The cache is per-group and per-table.

In order to disable the cache, one can set `skip_table_creation_cache`
to `true` in the module config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,5 +26,6 @@ before_install:
   - echo "-Xmx1g" >> ../apache-cassandra-${CASSANDRA_VERSION}/conf/jvm.options
   - sed -i -e 's/^-XX:+UseNUMA/#-XX:+UseNUMA/' ../apache-cassandra-${CASSANDRA_VERSION}/conf/jvm.options
   - bash ../apache-cassandra-${CASSANDRA_VERSION}/bin/cassandra > /dev/null
+  - sleep 10
 
 script: npm run-script coverage && (npm run-script coveralls || exit 0)
--- a/lib/db.js
+++ b/lib/db.js
@@ -3,6 +3,7 @@
 const P = require('bluebird');
 const cass = require('cassandra-driver');
 const extend = require('extend');
+const crypto = require('crypto');
 const dbu = require('./dbutils');
 const cassID = dbu.cassID;
 const SchemaMigrator = require('./schemaMigration');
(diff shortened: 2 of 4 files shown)


d00rman:
Travis seems to be having problems with Cassandra... ?

d00rman:
ok @Pchelolo this is now ready for your eyes :)

Pchelolo:
Did you try measuring the results of this? I honestly don't think this will make any difference all for the following reasons:

1. We don't actually create tables anymore in production, so there's no talking to Cassandra for realz.
2. All the table-create requests will be shuffled around backend nodes, so I am not really sure how much of a hit ratio will there be.

d00rman:
@Pchelolo on my laptop, it takes 50s for RB to start without this change and 30-35s with it (in both cases the tables were present in Cassandra). It's not revolutionary, but it does improve RB-backend's start-up time by 30%-40%, so I really think it's worth including at this stage.

Pchelolo:
> @Pchelolo on my laptop, it takes 50s for RB to start without this change and 30-35s with it (in both cases the tables were). It's not revolutionary, but it does improve RB-backend's start-up time by 30%-40%, so I really think it's worth including at this stage.

Do you have `skip_schema_update: true` enabled like in prod?

d00rman:
> Do you have `skip_schema_update: true` enabled like in prod?

I used the prod config, only changed the cassandra-related config (nodes, user, pass, ...), so yes, everything else is the same. To note that `skip_schema_update` only pertains to table schema migrations and replication settings, so it doesn't really make a difference for a stable config.

Pchelolo:
> > Do you have `skip_schema_update: true` enabled like in prod?
> 
> I used the prod config, only changed the cassandra-related config (nodes, user, pass, ...), so yes, everything else is the same. To note that `skip_schema_update` only pertains to table schema migrations and replication settings, so it doesn't really make a difference for a stable config.

This is just really surprising to me cause when nothing changed this patch seem to be adding work, not removing work. `this._migrateIfNessesary` should just return 201, and do nothing else. If all the schemas exist, `currentSchemaInfo` should be present at all times, so stuff on line 515 must never even be reached. So this patch should be adding a complicated digest calculation to hash a result of 1 method call that is supposed to return a constant. Hard to believe that would be a performance improvement. This means there's actually a bug somewhere. I think it would be nice to find that bug..

d00rman:
Oh hm yes, you are correct. In fact, I can't replicate the results I first reported. Now, I get similar times w/ and w/o this patch. My guess is Cassandra was acting up the first time around

